PR #37, opened 2023-02-28
Accordion Block

### LINKED ISSUE #6: Implement applications accordion block
- [ ] Desktop version (side by side accordion behaviour)
- [ ] Mobile version (classic accordion behaviour)

Desktop Version
Screenshot 2023-02-21 at 10 47 06

Mobile Version
Screenshot 2023-02-21 at 10 49 02

paolomoz:
Fix #6 

Test URLs:
- After: https://accordion-block--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/accordion/accordion.css
+++ b/blocks/accordion/accordion.css
@@ -2,15 +2,18 @@
 .accordion-tab-list {
     max-height: 800px;
 }
+
 .accordion-tab-main-content {
     background-color: #000;
     color: #fff;
 }
+
 .accordion-tab-main-content p {
     line-height: 1.5;
 }
+
 .accordion-tab-main-content h3 {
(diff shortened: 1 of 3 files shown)


lydiapuric (changes requested):
This is what I see what is different between the original and this branch:
- fonts don't match in size and weight
- the transition is too fast. the original one feels way smoother
- the space the text on the left side is too wide
- the 'learn more' button is too wide

Furthermore 
- IMO CSS !important should not be used, because the only thing that can override an !important tag is another !important tag. 
Instead please validate your CSS styles order (the order your styles are written). In general the best practice is to define the CSS selector as specific as possible. Sometimes we need to add CSS classes to make it specific enough.
- Please introduce and use the variables we have for fonts and colors.

PR #43, opened 2023-03-02
1 - Global Styling & Fonts

### LINKED ISSUE #1: Global Styling & Fonts
- [x] Headings
- [x] Margins
- [x] Custom Fonts
- [x] Font size
- [x] Colors
- [x] Buttons (primary, secondary)

andreituicu:

Fix #1

Test URLs:
- After: https://1-box-sizing--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -92,6 +92,10 @@
   --nav-height: 64px;
 }
 
+* {
+  box-sizing: border-box;
+}
+
 body {
   font-size: var(--body-font-size-m);
   margin: 0;


styles/styles.css
```diff
@@ -92,6 +92,10 @@
   --nav-height: 64px;
 }
 
+* {
```
mhaack, line 95: Is that the same on the original page, this is prettttty wide.

mhaack (approved):

styles/styles.css
andreituicu: I'm sorry Markus! I didn't understood your comment. Was that a question? :) 
Yes, it is in the original index page and checked a few others as well.

Screenshot 2023-03-02 at 13 02 24

PR #70, opened 2023-03-14
get in touch block created

### LINKED ISSUE #49: Implement get in touch
For reference see: https://www.moleculardevices.com/contact

image

kapildhiman:

Fix #49

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/contact
- After: https://get-in-touch--moleculardevices--hlxsites.hlx.page/contact

### DIFF
--- a/blocks/get-in-touch/get-in-touch.css
+++ b/blocks/get-in-touch/get-in-touch.css
@@ -1,0 +1,76 @@
+.get-in-touch > div {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.get-in-touch > div > div p {
+  margin-top: 0;
+}
+
+.get-in-touch > div > div h5 {
+  font-size: var(--heading-font-size-l);
+  font-weight: 600;
+  margin-bottom: 12px;
+  margin-top: 0;
+}
+
+.get-in-touch > div > div iframe {
+  border: 0;
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  display: block;
+  background: #fff;
+  padding-top: 10px;
+}
+
+.get-in-touch > div > div:last-child iframe {
+  height: 450px;
+}
+
+.get-in-touch > div > div .fa {
+  color: var(--text-green);
+  font-size: 36px;
+  padding-right: 20px;
+  float: left;
+  line-height: 25px;
+}
+
+.get-in-touch > div > div a {
+  color: inherit;
+}
+
+.hubspot-iframe-wrapper {
+  margin-left: -15px;
+  margin-right: -15px;
+}
+
+.hubspot-iframe-wrapper iframe {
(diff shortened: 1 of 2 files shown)


blocks/get-in-touch/get-in-touch.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.get-in-touch > div > div h5 {
+  font-size: var(--heading-font-size-l);
+  font-weight: 600;
+  margin-bottom: 12px;
+  margin-top: 0;
+}
+
+.hubspot-iframe-wrapper iframe {
```
mhaack, line 17: Prefix with `.get-in-touch`, we might have other hubspot frames with different hight? Or is this generic and can go to styles.css?

kapildhiman: That's a good point these forms are only related to get in touch so we are good with this height. 

Where we have different kinds of forms we can go with an iframe-resizer which will automatically calculate the height. 

Please Let me know if you have any other suggestions to handle this.

blocks/get-in-touch/get-in-touch.js
```diff
@@ -0,0 +1,26 @@
...
+  mapIframe.src = mapUrl.href;
+  mapIframe.setAttribute('loading', 'lazy');
+  mapUrl.parentNode.replaceChild(mapIframe, mapUrl);
+}
+
+export default function decorate(block) {
+  const observer = new IntersectionObserver((entries) => {
+    if (entries.some((e) => e.isIntersecting)) {
+      observer.disconnect();
+      addIframe();
```
andreituicu, line 22: Usually I would say to pass the `block` as a parameter to the function, so all query selector interrogations are relative just to the block, in case in the future, someone adds the component more than once in the page.

In this particular case, I guess nobody adds this component more than once, so shouldn't be a problem.

andreituicu (approved):

blocks/get-in-touch/get-in-touch.css
mhaack: If it is only for this block we should have it scoped for the block only.

Seen the iframe-resizer on the current page, I would like to avoid using it if possible. Automatic resizing can add to CLS.

blocks/get-in-touch/get-in-touch.js
mhaack: +1 good catch.

jose-correia (approved):

mhaack (approved):

blocks/get-in-touch/get-in-touch.css
kapildhiman: css class name updated
.
Sure will target CLS for the iframe.

PR #88, opened 2023-03-15
17 - Wave section carousel

### LINKED ISSUE #17: Implement "wave" section
- [ ] implement using section metadata with style and background
- [ ] content should be centered
- [ ] different styling for primary and secondary buttons
- [ ] most work with different content & blocks

Screenshot 2023-02-21 at 13 42 36

Screenshot 2023-02-21 at 13 51 21

Screenshot 2023-02-21 at 13 59 38

Example:
* https://www.moleculardevices.com/products/microplate-readers
* https://www.moleculardevices.com/service-support/lab-automation-solutions#features
* https://www.moleculardevices.com/applications
* https://www.moleculardevices.com/service-support
* https://www.moleculardevices.com/products/clone-screening

andreituicu:

Fix #17 

Test URLs:
- Before: https://17-wave-section--moleculardevices--hlxsites.hlx.page/
- After: https://17-wave-carousel2--moleculardevices--hlxsites.hlx.page/

- Before: https://17-wave-section--moleculardevices--hlxsites.hlx.page/drafts/tuicu/wave-carousel
- After: https://17-wave-carousel2--moleculardevices--hlxsites.hlx.page/drafts/tuicu/wave-carousel


styles/styles.css
```diff
@@ -391,7 +391,7 @@ main .section>div {
 /* progressive section appearance */
 main .section[data-section-status='loading'],
 main .section[data-section-status='initialized'] {
-  display: none;
+  display: none !important;
```
mhaack, line 394: Why is this needed?

scripts/lib-franklin.js
```diff
@@ -283,7 +283,7 @@ export function decorateSections(main) {
           styles.forEach((style) => section.classList.add(style));
         } else if (key === 'background') {
           const { background } = meta;
-          if (background.startsWith('https://')) {
+          if (background.startsWith('http')) {
```
mhaack, line 286: Why? We should have all URLs with `https`

andreituicu: Yes, except when working locally with `hlx up` :) There we have `http://localhost...`, so the background is not displayed.

mhaack: Ok I see.

jose-correia:
@andreituicu can't we add the quotes to the content document, instead of adding them to the source code? or does it increase the complexity by a lot?

andreituicu:
@jose-correia in theory we could, as far as I understand it doesn't give us too many advantages.
We would assume that if we find 2 images in the text, those are qoutes and style accordingly. This gives at most the ability to authors to use other quote pictures in the future.
However, as disadvantages, it would make our life harder in the future if we want to have other images in text (like the authors picture for example next to their name), because then it would be more complex to figure out which picture is the authors and which pictures are the quotes.

With the current implementation that Markus proposed, we can use qoutes in any other carousels by simply putting the `blockquote` styling on the text you want the qoutes to be present and other images can have different semantics.

This is my understanding. I hope it helps! :)

PR #117, opened 2023-03-25
Add navbar dropdown menus (left-side only)

jose-correia:
This PR adds the dropdown logic and the left-side part of the dropdown menus.

The dropdowns are loaded on the background after loading the header so that we don't impact the Lighthouse score with the files that need to be fetched.

This PR also separates the header logic into multiple files so that it's easier to navigate and improves code reuse

Fix #<gh-issue-id>

Test URLs:
- After: https://navbar-products-megamenu--moleculardevices--hlxsites.hlx.page/drafts/josec/


blocks/header/header.css
```diff
@@ -158,7 +170,96 @@ header nav .nav-tools {
...
+  padding: 0;
+  overflow: hidden;
+  z-index: 0;
+  margin: 0 !important;
+  display: block;
+  opacity: 1;
+  visibility: visible;
+  position: absolute;
+  background: #fff url('/images/header-background-filler.jpg') no-repeat right bottom;
+  background-repeat: no-repeat;
```
mhaack, line 189: Line above already has `no-repeat` this one should not be needed.

blocks/header/header.css
```diff
@@ -158,7 +170,96 @@ header nav .nav-tools {
...
+}
+
+.header .nav-tabs .head-submenu {
+  display: flex;
+  justify-content: space-between;
+}
+
+.header .nav-tabs .head-submenu > div {
+  background-size: 250px;
+  color: #fff;
```
mhaack, line 241: Use color variables.

blocks/header/header.css
```diff
@@ -158,7 +170,96 @@ header nav .nav-tools {
...
+  width: 50%;
+}
+
+header .nav-tabs .menu-nav-submenu ul {
+  float: none;
+}
+
+header .nav-tabs .menu-nav-submenu-sections {
+  display: block;
+  border-right: 1px solid #FFFFFF80;
```
mhaack, line 256: Use color variables.

scripts/scripts.js
```diff
@@ -112,6 +112,26 @@ export function addFavIcon(href, rel = 'icon') {
...
 }
 
+/**
+ * Takes a string as an input, converts it to lowercase,
+ * replaces special characters with hyphens,
+ * and removes any trailing or leading hyphens.
+ * @param {*} inputString
+ * @returns string
+ */
+export function getIdFromString(inputString) {
```
mhaack, line 122: We already have `toClassName` helper? Can't this be used here as well instead of new helper?

jose-correia: Nice! Didnt know about that one

blocks/header/header-events.js
```diff
@@ -0,0 +1,68 @@
...
+  if (mediaQueryList.matches) {
+    addEventListenersDesktop();
+  } else {
+    addEventListenersMobile();
+  }
+}
+
+export function handleViewportChanges(block) {
+  mediaQueryList.onchange = () => {
+    document.querySelector('main').style.visibility = '';
```
andreituicu, line 63: I don't fully understand this change. why do we need to update the visibility of `main`?

andreituicu (approved):
LGTM!

Small comments above.

blocks/header/header-events.js
```diff
@@ -0,0 +1,68 @@
...
+    el.replaceWith(el.cloneNode(true));
+  });
+  elementsWithEventListener = [];
+}
+
+function addEventListenersDesktop() {
+  function expandMenu(element) {
+    const expanded = element.getAttribute('aria-expanded') === 'true';
+    collapseAllSubmenus(element.closest('ul'));
+    element.setAttribute('aria-expanded', expanded ? 'false' : 'true');
```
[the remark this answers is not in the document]
mhaack: Not sure, not an expert here but I think setting `aria-expanded` to false is valid. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

scripts/scripts.js
```diff
@@ -117,7 +119,12 @@ export function addFavIcon(href, rel = 'icon') {
  */
 async function loadLazy(doc) {
   const main = doc.querySelector('main');
+
   loadHeader(doc.querySelector('header'));
+
+  // Fetch and style the megamenus in the background
+  fetchAndStyleHeaderMegamenus(doc.querySelector('header'));
```
mhaack, line 126: Thinking if we need this at all? Already `loadHeader` is called lazy, can't we merge it with that function?

blocks/header/header-events.js
jose-correia: Good point, this logic was copied from the another Franklin block, deleting doesn't seem to have any impact at first, will remove

PR #130, opened 2023-03-29
Add navbar products header (right component)

jose-correia:

Screenshot 2023-03-29 at 18 33 53

Test URLs:
- After: https://navbar-add-products-header--moleculardevices--hlxsites.hlx.live/


mhaack:
The teaser cards should have a link, then entire card, not only the title.

jose-correia:
@mhaack Fixed 👍 The behavior is now the same as the original website. The card is not a link, but the green text and the image are.

mhaack (approved):

PR #140, opened 2023-03-30
Fix Template Loading

andreituicu:

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/newsroom/news/advanced-solutions-life-sciences-collaborate-to-develop-3d-biology-automation-technologies-for-drug-discovery
- After: https://fix-template-load--moleculardevices--hlxsites.hlx.page/newsroom/news/advanced-solutions-life-sciences-collaborate-to-develop-3d-biology-automation-technologies-for-drug-discovery

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -118,11 +118,11 @@ async function decorateTemplates(main) {
  * @param {Element} main The main element
  */
 // eslint-disable-next-line import/prefer-default-export
-export function decorateMain(main) {
+export async function decorateMain(main) {
   // hopefully forward compatible button decoration
   decorateButtons(main);
   decorateIcons(main);
-  decorateTemplates(main);
+  await decorateTemplates(main);
   decorateSections(main);
   decoratePageNav(main);
   decorateBlocks(main);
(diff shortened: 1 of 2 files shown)


scripts/scripts.js
```diff
@@ -118,11 +118,11 @@ async function decorateTemplates(main) {
  * @param {Element} main The main element
  */
 // eslint-disable-next-line import/prefer-default-export
-export function decorateMain(main) {
+export async function decorateMain(main) {
   // hopefully forward compatible button decoration
   decorateButtons(main);
   decorateIcons(main);
-  decorateTemplates(main);
+  await decorateTemplates(main);
```
mhaack, line 125: I wounder if we should move this call to `loadEager`. Checked in debugger this is called to often if we have fragments. Since we call `decorateMain` in fragement.js as well. But `decorateTemplates` should only be called once per page load.

andreituicu: In this case, fragments that are included in pages with autoblocking or any other template logic would not apply to fragments. Even if you have a fragment page that has specifically a `template` metadata declared.
So looking at a fragment:
* as a standalone page could look one way
* when included in a different page, could look different

But that obviously depends if what's in that specific fragment

mhaack: > fragments that are included in pages with autoblocking or any other template logic would not apply to fragments
If we call `decorateTemplates` before `decorateMain` it should work. All template specific stuff we run before `decorateMain` kicks in.

mhaack (approved):

PR #147, opened 2023-04-03
Add navbar text component

jose-correia:

Screenshot 2023-04-03 at 12 23 58

Fix #<gh-issue-id>

Test URLs:
- After: https://header-text-component--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -473,7 +473,7 @@ header .cards-submenu p {
   margin-top: 0;
 }
 
-header .cards-submenu p:has(picture) {
+header .cards-submenu p:nth-child(1) {
   background-color: var(--text-white);
   min-height: 114px;
   overflow: hidden;
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -567,12 +567,24 @@ header .cards-submenu ul li a:hover {
     display: block;
   }
 
-  header .cards-submenu p:nth-child(2) a {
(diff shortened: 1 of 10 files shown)


blocks/header/header.css
```diff
@@ -51,6 +51,35 @@ header .navbar-header > .company-links {
...
+  justify-content: space-between;
+}
+
+header .text-submenu p {
+  min-height: auto;
+  color: #fff;
+  padding-left: 0;
+  text-transform: uppercase;
+  font-size: 15px;
+  font-family: "Proxima Nova Regular", sans-serif;
```
mhaack, line 66: Do we need to set font explicitly on `p`? If yes please use variable, so we have the web font and the fall back font.

blocks/header/header.css
```diff
@@ -314,6 +352,20 @@ header .nav-tabs .menu-nav-submenu h1:hover {
   cursor: pointer;
 }
 
+header .text-submenu ul > li > a {
+  font-size: 11px;
+  color: #fff;
+  border-bottom: 0;
+  font-family: "Proxima Nova Regular", sans-serif;
```
mhaack, line 359: Same as above, do we need to set it explicitly?

mhaack (approved):

blocks/header/header.css
jose-correia: We dont, nice catch

Issue #175, opened 2023-04-06
Coveo behavour for support.moleculardevices.com users


kapildhiman:
Show coveo results according to the role of the logged-in users in support.moleculardevices.com users

![image](https://user-images.githubusercontent.com/58678719/230348113-b00e4217-9e2b-4174-b67f-8109b7a0dab2.png)

mhaack:
@kapildhiman this was implemented right? Anything left over?

PR #181, opened 2023-04-06
coveo search block

### LINKED ISSUE #116: Site Search Coveo
https://www.moleculardevices.com/search-results#t=All&sort=relevancy

image

kapildhiman:

Fix #116, #175

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/search-results
- After: https://coveo--moleculardevices--hlxsites.hlx.page/search-results


scripts/scripts.js
```diff
@@ -268,6 +268,52 @@ export function getQueryParameter() {
...
+  date.setTime(date.getTime() + (exdays * 24 * 60 * 60 * 1000));
+  if (exdays !== 0) {
+    expires = `expires=${date.toUTCString()}`;
+  }
+
+  domain = window.location.hostname.endsWith('.moleculardevices.com');
+  if (domain === true) {
+    hostName = 'domain=.moleculardevices.com;';
+  }
+  /* eslint-disable no-trailing-spaces */
```
mhaack, line 291: This can be easily fixed and should not be silenced.

scripts/scripts.js
```diff
@@ -268,6 +268,52 @@ export function getQueryParameter() {
...
+  document.cookie = `${cname}=${cvalue};secure;${hostName}${expires};path=/`;
+}
+
+/**
+ * Get a cookie
+ * @param cname the name of the cookie
+ */
+export function getCookie(cname) {
+  /* eslint-disable no-param-reassign */
+  cname += '=';
```
mhaack, line 301: this should also not be silenced and can be fixed by `const cname = "${cname}=";`

blocks/cp-request/cp-request.js
```diff
@@ -0,0 +1,13 @@
+import { getQueryParameter, setCookie } from '../../scripts/scripts.js';
+
+const paramName = 'STYXKEY_PortalUserRole';
+const readQuery = getQueryParameter();
+const queryStringParam = readQuery[paramName] ? readQuery[paramName] : '';
+
+if (queryStringParam && (document.referrer.includes('.moleculardevices.com') || document.referrer.includes('hlxsites.hlx.') || document.referrer.includes('localhost:3000'))) {
```
mhaack, line 7: Better check for `.hlx.live` and `.hlx.page` instead of hlxsites.hlx. As if the repo gets moved into some other org this will break.

blocks/coveo-search/coveo-search.js
```diff
@@ -0,0 +1,270 @@
...
+  } else {
+    filter = `NOT @sfkbid OR (  @sfdatacategoryaccess_level == (${accessLevel})  OR @sfisvisibleinpkb=true  )`;
+  }
+  return filter;
+}
+
+function searchFormHeader() {
+  return `
+          <div id="search" class="CoveoSearchInterface mdcoveo" data-enable-history="true" data-excerpt-length="350">
+            <div class="CoveoFolding"></div>
```
mhaack, line 59: Here are many class names in camel case like `CoveoFolding` which are not styled in the css file. Is that copy & pages and can be removed?

scripts/scripts.js
kapildhiman: This has been fixed.

blocks/cp-request/cp-request.js
kapildhiman: Added .hlx.live and .hlx.page in condition.

blocks/coveo-search/coveo-search.js
kapildhiman: Most of them are used. 

image

e.g. CoveoAnalytics is used for coveo tracking. We should avoid removing there could be specific uses of these in coveo.

mhaack (approved):

kapildhiman:
Thanks, @paolomoz  for the suggestion. I've made the changes as per your suggestion.

PR #187, opened 2023-04-11
index h1 for breadcrumbs title

paolomoz:
Breadcrumbs title is using the H1 value hence it needs to be indexed to be retrieved for current page and ancestor pages at:
https://github.com/hlxsites/moleculardevices/blob/27950d2c8ce413166ae269d7415c1a8e0f41612f/blocks/breadcrumbs/breadcrumbs-create.js#L31

### DIFF
--- a/helix-query.yaml
+++ b/helix-query.yaml
@@ -54,6 +54,9 @@ indices:
       lastModified:
         select: none
         value: parseTimestamp(headers["last-modified"], "ddd, DD MMM YYYY hh:mm:ss GMT")
+      breadcrumbsTitle:
+        select: h1
+        value: textContent(el)
   fragments:
     include:
       - '/fragments/**'
--- a/helix-query.yaml
+++ b/helix-query.yaml
@@ -54,7 +54,7 @@ indices:
       lastModified:
         select: none
         value: parseTimestamp(headers["last-modified"], "ddd, DD MMM YYYY hh:mm:ss GMT")
-      breadcrumbsTitle:
+      h1:
         select: h1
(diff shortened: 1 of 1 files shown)


andreituicu (approved):

helix-query.yaml
```diff
@@ -54,6 +54,9 @@ indices:
       lastModified:
         select: none
         value: parseTimestamp(headers["last-modified"], "ddd, DD MMM YYYY hh:mm:ss GMT")
+      breadcrumbsTitle:
```
andreituicu, line 57: Maybe don't make it so specific? something like `heroHeader`, `mainHeader`, `header1`? It could be used for other purposes, other than breadcrumbs.

mhaack: I suggest to index stuff for breadcrumbs into a separate sheet and index to not pollute the main index with soo many special purpose fields.

paolomoz: @andreituicu @mhaack thanks. So, for now this h1 value is the only special property we need. I suggest to keep it in the main index (for now) also because it is anyway needed for most pages and because we would have a separate index for just 1 property. I renamed it to 'h1' so it is what it is, independent from the use we make of it. wdyt?

andreituicu: @paolomoz @mhaack : I tend to agree with @paolomoz in this instance. I've seen other places where the `h1` is used instead of the title (e.g. the recent blogs carousel, might be other) and it would make rendering slower for the breadcrumb if we need to make 2 get requests to two different indexes, which is present in a lot of pages.

PR #194, opened 2023-04-12
improve LCP

paolomoz:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/cellular-imaging-systems/high-content-imaging/imagexpress-micro-confocal
- After: https://improve-lcp--moleculardevices--hlxsites.hlx.page/products/cellular-imaging-systems/high-content-imaging/imagexpress-micro-confocal

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -18,7 +18,7 @@ import {
 import loadHeader from './header-utils.js';
 import TEMPLATE_LIST from '../templates/config.js';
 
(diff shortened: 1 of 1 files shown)


lydiapuric (approved):
+1
The hero block contains the image that is the first impression on a page and should not be lazy loaded.
Blocking the LCP til the hero component is loaded makes sense to me.

paolomoz:
closing (might reopen when we get back to improving)

PR #222, opened 2023-04-19
Implement Citations list page

### LINKED ISSUE #96: [Citations] Implement Citations list page & block
Reference: https://www.moleculardevices.com/citations

Data: https://main--moleculardevices--hlxsites.hlx.live/resources/citations/query-index.json?sheet=citations

jose-correia:

Fix #96

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/citations
- After: https://citations-list-page--moleculardevices--hlxsites.hlx.page/citations

### DIFF
--- a/blocks/citation-list/citation-list.css
+++ b/blocks/citation-list/citation-list.css
@@ -1,5 +1,5 @@
 main .citation-list-container {
-  background-image: url(/images/grey_molecules.png);
+  background-image: url("/images/grey_molecules.png");
   background-repeat: no-repeat;
   background-position: center top;
   background-size: auto 90px;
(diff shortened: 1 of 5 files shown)


blocks/citation-list/citation-list.css
```diff
@@ -0,0 +1,185 @@
+main .citation-list-container {
+  background-image: url("/images/grey_molecules.png");
```
mhaack, line 2: This should be come from the section, not the block itself.

mhaack (changes requested):

blocks/citation-list/citation-list.css
mhaack: And we have https://github.com/hlxsites/moleculardevices/blob/main/styles/styles.css#L504 for that, might be broken by @andreituicu hero changes as well.

mhaack: Here is the fix for the molecules bg: #230 230

PR #234, opened 2023-04-20
Fetch WaveCarousel background from code instead of doc

### LINKED ISSUE #228: Adapt "WaveCarousel" sections for blue and green background from css
Wave sections `WaveCarousel, Green` & `WaveCarousel, Blue` currently get the background from document but should get it from css & git code.

jose-correia:

Fix #228

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/drafts/tuicu/wave-carousel
- After: https://fetch-wave-background-from-code--moleculardevices--hlxsites.hlx.page/drafts/josec/wave-carousel


styles/styles.css
```diff
@@ -784,6 +784,14 @@ main .section.wave.wavecarousel {
   background-size: 100% 100%;
 }
 
+main .section.wave.wavecarousel.blue {
+  background-image: url("/images/wave-background-blue.webp");
+}
+
+main .section.wave.wavecarousel.bluegreen {
+  background-image: url("/images/wave-background-blue-green.webp");
```
andreituicu, line 792: I'm still not sure if it is safe to have the `webp` only, without the `png` too. As in:
`background-image: url("/images/wave-background-blue-green.webp"), url("/images/wave-background-blue-green.png");`

Everywhere else in the code, we keep compatibility for browsers that don't support yet `webp`.

andreituicu (approved):
LGTM! Only the question above.

styles/styles.css
jose-correia: Good point! added the png versions

PR #238, opened 2023-04-20
chore: implement cards carousel

mhaack:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/drafts/mhaack/carousel
- After: https://carousel-cards--moleculardevices--hlxsites.hlx.page/drafts/mhaack/carousel

### DIFF
--- a/blocks/carousel/carousel.css
+++ b/blocks/carousel/carousel.css
@@ -1,3 +1,5 @@
+@import url("carousel-cards.css");
+
 /* 
   Default Carousel Styling. 
   Same as .products on the landing page.
--- a/blocks/carousel/carousel.js
+++ b/blocks/carousel/carousel.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-unused-expressions */
 import { decorateIcons, loadCSS } from '../../scripts/lib-franklin.js';
+import { span } from '../../scripts/dom-helpers.js';
 
 const AUTOSCROLL_INTERVAL = 7000;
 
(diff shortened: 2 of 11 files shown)


mhaack:
@andreituicu I change `export async function createCarousel` (removed default) so we can actually use the carousel directly. With that we can maybe also change "feature-products-carousel" to "carousel" and remove one extra block. WDYT?

lydiapuric (approved):
LGTM +1

blocks/carousel/carousel.css
```diff
@@ -1,3 +1,5 @@
+@import url("carousel-cards.css");
```
andreituicu, line 1: I think we should either use the `loadCSS` function for this or inline the `carousel-cards.css`.
Otherwise we have the worst of both worlds right now: All pages that don't need the `carousel-cards.css` still load it, but via an extra http request.

blocks/carousel/carousel.js
```diff
@@ -191,7 +179,7 @@ class Carousel {
     const buttonLeft = document.createElement('button');
     buttonLeft.classList.add('carousel-nav-left');
     buttonLeft.ariaLabel = 'Scroll to previous item';
-    createNavButtonIcon(buttonLeft, 'left');
+    buttonLeft.append(span({ class: 'icon icon-chevron-left' }));
```
andreituicu, line 182: ❤️

blocks/carousel/carousel.css
mhaack: Ok will fix that

blocks/carousel/carousel.js
mhaack: We could do the button as well but maybe let’s do that as separate optimization PR.

blocks/carousel/carousel.css
mhaack: @andreituicu please check again, loaded via `loadCSS` now, with a common loader for all css default plus extra css files.

andreituicu:
@mhaack Since the `defaultStyling` carousel configuration was changed, these too need to be adapted:

https://github.com/hlxsites/moleculardevices/blob/main/blocks/citation-list/citation-list.js#L63
https://github.com/hlxsites/moleculardevices/blob/main/blocks/blog-list/blog-list.js#L47

afterwards should be good to go.

andreituicu (approved):

PR #244, opened 2023-04-21
feat: add share story block

### LINKED ISSUE #168: [Customer Breakthrough] Implement "Share Story" block
- only middle c2a text is in block the left & right columns are dynamic
- share buttons have dynamic links for the current page
- "Submit your Story" opens lightbox with Hubspot form - check with MolDev Team

Sample 1: https://main--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/br/ionic-transport-assays
Reference 2: https://www.moleculardevices.com/en/assets/customer-breakthrough/br/ionic-transport-assays

duynguyen:
Fix #168 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/dd/img/visikol
- After: https://168-share-story--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/dd/img/visikol


blocks/share-story/share-story.js
```diff
@@ -0,0 +1,96 @@
...
+      ul({ class: 'button-container' },
+        ...socials.map((social) =>
+        // eslint-disable-next-line implicit-arrow-linebreak
+          li({ class: `share-${social}`, 'data-type': social },
+            i({ class: `fa fa-${social}` }),
+          ),
+        ),
+      ),
+    ),
+  );
```
andreituicu, line 45: not blocking: is this copied from `social-share.js`? would it make sense to make it as a common function?
( just a question 🙂 don't need to do it if you don't think it's worth it)

e.g.
```
export function socialShareBlock(socials) {
  return (
    div({ class: 'share-event' },
      p('share this story'),
      div({ class: 'social-links' },
        ul({ class: 'button-container' },
          ...socials.map((social) =>
          // eslint-disable-next-line implicit-arrow-linebreak
            li({ class: `share-${social}`, 'data-type': social },
              i({ class: `fa fa-${social}` }),
            ),
          ),
        ),
      ),
    )
  )
}

andreituicu (approved):
Code looks good!

@mhaack I see there's a problem with the `og:image`. It doesn't look the same when trying to share to facebook. In the original the product picture is used, here it is the hero. We might need to add that as metadata.

mhaack:
Need to check normally the import should capture the og:image as well. Hero or first image of content is automatically used if no og:image is there. Like for this doc, it has no image in meta data. So yes might be an important bug.

blocks/share-story/share-story.js
duynguyen: good point 👍 I've just added

PR #258, opened 2023-04-24
cmp function added in scripts.js which is requred for internal forms.

### LINKED ISSUE #257: Set a cmp session cookie if cmp exists in query string
e.g. 

image

This cookie will be required for RFQ forms Landing pages forms etc.

kapildhiman:

Fix #257

Test URLs:
- After: https://cmp--moleculardevices--hlxsites.hlx.page/?cmp=7014u000001AMomAAG

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -407,13 +407,12 @@ export function getCookie(cname) {
 }
 
 /**
- * Set a cmp session cookie if cmp exits in query string
+ * Set a cookie from query string parameters
  */
-function setCmpCookie() {
+function setCookieFromQueryParameters(paramName, exdays) {
   const readQuery = getQueryParameter();
-  const paramName = 'cmp';
   if (readQuery[paramName]) {
-    setCookie(paramName, readQuery[paramName], 0);
+    setCookie(paramName, readQuery[paramName], exdays);
   }
 }
 
@@ -423,5 +422,5 @@ async function loadPage() {
   loadDelayed();
 }
 
-setCmpCookie();
+setCookieFromQueryParameters('cmp', 0);
 loadPage();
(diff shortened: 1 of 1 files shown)


scripts/scripts.js
```diff
@@ -406,10 +406,22 @@ export function getCookie(cname) {
...
+  }
+}
+
 async function loadPage() {
   await loadEager(document);
   await loadLazy(document);
   loadDelayed();
 }
 
+setCmpCookie();
```
jose-correia, line 426: Isn't it possible that we may need to set other cookies from parameters? Could we benefit from the extensibility of having instead something like this?

```javascript
function setCookieFromQueryParameters(paramName) {
  const readQuery = getQueryParameter();
  if (readQuery[paramName]) {
    setCookie(paramName, readQuery[paramName], 0);
  }
}

setCookieFromQueryParameters('cmp');
```

jose-correia (changes requested):

scripts/scripts.js
```diff
@@ -406,10 +406,22 @@ export function getCookie(cname) {
   return '';
 }
 
+/**
+ * Set a cmp session cookie if cmp exits in query string
+ */
+function setCmpCookie() {
+  const readQuery = getQueryParameter();
```
jose-correia, line 413: This function should be called `getQueryParameters` right?

jose-correia (changes requested):

scripts/scripts.js
kapildhiman: So far I am only seeing the cmp cookie as a session cookie, but it's a good suggestion @jose-correia. Let's keep it this way I'll make some other changes.

jose-correia (approved):

PR #260, opened 2023-04-24
feat: implement featured products carousel mini

### LINKED ISSUE #166: [Customer Breakthrough] Implement `Featured Products Carousel (mini)` block version
Implement the mini version of the featured products carousel.

- carousel can have 1 or multiple entries
- all content is in the word document (image column, text & link column)
- blue parallax background is implemented via #167

Sample 1: https://main--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/br/ionic-transport-assays
Reference 2: https://www.moleculardevices.com/en/assets/customer-breakthrough/br/ionic-transport-assays

Sampel 2: https://main--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/dd/img/gustave-roussy-cancer-center
Reference 2: https://www.moleculardevices.com/en/assets/customer-breakthrough/dd/img/gustave-roussy-cancer-center

duynguyen:
Fix #166 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/dd/img/visikol
- After: https://166-products-carousel-mini--moleculardevices--hlxsites.hlx.page/en/assets/customer-breakthrough/dd/img/visikol
- Original: https://www.moleculardevices.com/en/assets/customer-breakthrough/dd/img/visikol

### DIFF
--- a/blocks/carousel/carousel.js
+++ b/blocks/carousel/carousel.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-unused-expressions */
 import { decorateIcons, loadCSS } from '../../scripts/lib-franklin.js';
-import { span } from '../../scripts/dom-helpers.js';
+import { div, span, p } from '../../scripts/dom-helpers.js';
 
 const AUTOSCROLL_INTERVAL = 7000;
 
@@ -24,9 +24,13 @@ class Carousel {
     this.defaultStyling = false;
     this.dotButtons = true;
     this.navButtons = true;
+    this.counter = false;
     this.infiniteScroll = true;
     this.autoScroll = true; // only available with infinite scroll
     this.autoScrollInterval = AUTOSCROLL_INTERVAL;
+    this.currentIndex = 0;
+    this.counterText = '';
+    this.counterNavButtons = true;
     // this is primarily controlled by CSS,
     // but we need to know then intention for scrolling pourposes
     this.visibleItems = [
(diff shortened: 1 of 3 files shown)


blocks/carousel/carousel.js
```diff
@@ -377,7 +409,8 @@ class Carousel {
     this.autoScroll && this.infiniteScroll
       && (this.intervalId = setInterval(() => { this.nextItem(); }, this.autoScrollInterval));
     this.dotButtons && this.createDotButtons();
-    this.navButtons && this.createNavButtons();
+    this.counter && this.createCounter();
```
mhaack, line 412: I wounder if we can have the counter & text as an alternative to the nav buttons? Idea: the featured products block just brings its own `createNavButtons();` function which renders the counter and text as needed instead of adding this as a second option?
@andreituicu any thoughts on this?

duynguyen: At first glance it sounds logical to pass custom `createNavButtons()`. However, the more I think about it, the more problematic it might become:
- We need to either duplicate lots of OOTB nav logics in the custom function, or expose quite a few internal Carousel class behaviors so that the external custom function can call.
- Because the internal of `createNavButtons` also needs to update back to the "custom" HTML (updating page counter), it gets even more complex to refactor the carousel.

If adding another option for `counter` is not preferred, we could add the counter to the default rendered div and control it by CSS. However it could make performance a bit worse as you always have to update the counter on each nav behavior anyway (regardless it's visible or hidden).

If we have another option for `counter`, we can simplify it a bit by:
- if it is false: no counter is added
- if it is true: default counter is added, e.g. "1 / 1"
- if it is a string "Product", counter is added with the text, e.g. "Product 1 / 1"
We will also not need `counterNavButtons`. That depends on the single `navButtons` option. Hence we only add the `counter` option in this PR.

@mhaack @andreituicu WDYT?

andreituicu: I kind of agree with @duynguyen , on this one. I agree that it would be very nice to allow consumers to provide a `createNavButtons` custom function it would need quick some refactoring to make it work to attach the event listeners. @mhaack Do you have examples from the website where this would be helpful? I'm having a bit of a problem to picture when would we need it.

BTW, anything can be custom in the carousel intentionally, because of: https://github.com/hlxsites/moleculardevices/pull/260/files#diff-0ba135df53c420d893ac5474ae71abe07027087aba7d760a4534f1d585cefd4bL48 . As long as you put a function with the same name as the one from the class in the config object.

andreituicu (approved):
LGTM!

Minor (not blocking from my personal point of view):
* There is no gap/margin between carousel items on the original website for the mini version
* The details button icon has the colors reversed.

blocks/carousel/carousel.js
mhaack: I think we are good here. Thanks for sharing the details.

mhaack (approved):

PR #271, opened 2023-04-27
Add mobile menu items and search

jose-correia:
This PR adds the first mobile menu view that is seen after clicking the hamburger button.

Keep in mind that each of the subcategories doesn't work yet. Since this PR was becoming quite big I decided to split the changes.

## Changes

- Refactored some of the code so that it uses the dom-helpers
- Refactored some of the code so that the mobile menu's subcategories could be built along the desktop megamenu. This reduces the number of times we poll the content and iterate over it
- Added code to build the mobile menu list items and style it

## Notes

- The subcategories don't work yet, so clicking the arrows shouldn't do anything
- Submitting the search form results in the usage of the wrong parameters. Will be fixed in the next PR

## Look
Screenshot 2023-04-27 at 15 18 52

Test URLs:
- After: https://add-mobile-menu-initial-screen--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/header/header-events.js
+++ b/blocks/header/header-events.js
@@ -76,7 +76,13 @@ function addEventListenersDesktop() {
   addListeners('.search-form', 'submit', (e) => {
     e.preventDefault();
     e.stopPropagation();
-    submitSearchForm(e);
+    submitSearchForm(e, 'searchQuery');
+  });
+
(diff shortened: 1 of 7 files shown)


lydiapuric (approved):
Looks good to me +1

PR #272, opened 2023-04-27
feat: implement resources carousel

### LINKED ISSUE #224: [Products] Implement related resources carousel
* Block renders a carousel with 8 resources
* Resource sheet to be used https://main--moleculardevices--hlxsites.hlx.page/query-index.json
* Resources are filtered by product or application name based page it is used on

Sample: https://main--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-m-series-readers look for "Resources Carousel" block in the related resources area.

duynguyen:
Fix #224 

Test URLs:
- After: https://224-resources-carousel--moleculardevices--hlxsites.hlx.page/products/clone-screening/mammalian-screening/cloneselect-imager
- Original: https://www.moleculardevices.com/products/clone-screening/mammalian-screening/cloneselect-imager

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/applications/3d-cell-models/organoids/intestinal-organoids
- After: https://224-resources-carousel--moleculardevices--hlxsites.hlx.page/applications/3d-cell-models/organoids/intestinal-organoids
- Original: https://www.moleculardevices.com/applications/3d-cell-models/organoids/intestinal-organoids

### DIFF
--- a/blocks/applications-carousel/applications-carousel.css
+++ b/blocks/applications-carousel/applications-carousel.css
@@ -36,6 +36,7 @@ main .carousel.applications-carousel {
 .carousel.applications-carousel .card-description {
   font-size: var(--body-font-size-xs);
   color: var(--text-color);
+  margin-bottom: 10px;
 }
 
 .carousel.applications-carousel .card .button-container {


lydiapuric:
@duynguyen The cutoff of description looks pretty weird, but I know this is the text you got as content.
The image is broken in mobile view.

mhaack:
@duynguyen can we filter out the ‘0’ on description and show nothing in that case. Also if no image we should render a default placeholder let’s use the blue default header we have for some docs. 

There can always be the case author does not enter a description or forget the image. This should not break the carousel.

blocks/resources-carousel/resources-carousel.js
```diff
@@ -0,0 +1,48 @@
+import ffetch from '../../scripts/ffetch.js';
+import { getMetadata } from '../../scripts/lib-franklin.js';
+import { renderBlogCard } from '../../templates/blog/blog.js';
```
mhaack, line 3: Let’s rename this card to a generic name and move it to the carousel.js. WDYT.

duynguyen: yes, I would do that. @andreituicu good for you?

mhaack:
Perfect let’s use that image.

mhaack:
The default image fallback should go into https://github.com/hlxsites/moleculardevices/blob/f6ea4bda64533659bce066e1455af487a97f59a5/templates/blog/blog.js#L54 (or wherever we move the render function) so it work for all carousel using the render function.

lydiapuric:
The code looks good to me.

Regarding functionality we have a critical gap in Mobile:
- The swiping in Mobile does not work for me. 

Other minor observations:
- I see only 1 card in Mobile, no indication that there are more cards.
- The swiping of the carousel is currently implemented as card by card. On the original page all 3 cards are exchanged.
- The button text in original page is 'Learn more' instead of 'Read More'

andreituicu:
> The swiping in Mobile does not work for me. (By resizing window only, without Mobile view, this issues are not reproducable in FF for me, but in Chrome still)

Swipping on Mobile is implemented with touch ui events, so on Chrome it needs Mobile view to be activated for it to work. It is indeed different than on the original website, but on Mobile devices (at least on my phone - Samsung) swiping works, haven't tried on an iPhone. So the only thing that is missing is on Desktops with very small window sizes for browsers, but  I'm not sure if that is an usecase for users.

lydiapuric:
@andreituicu Are you able to swipe the cards in Chrome, Mobile view enabled? - because I am not

andreituicu:
@lydiapuric Yes, it does for me. Sharing a video on Slack

lydiapuric:
Discussed with @andreituicu. It is expected that swiping does not work in Chrome, mobile view. Instead touch works.

duynguyen:
@lydiapuric - updated the PR according to your reviews:

- I see only 1 card in Mobile, no indication that there are more cards
- The button text in original page is 'Learn more' instead of 'Read More'

I would defer this for now as it impacts global carousel config. Proposing we park it for post-live:

- The swiping of the carousel is currently implemented as card by card. On the original page all 3 cards are exchanged.

WDYT?

mhaack (approved):

Issue #219, opened 2023-04-19
Implement generic video URL transforming for all video links


mhaack:
To use video links across the content, als within column block for example we should change from block to generic video links parsing and transformation for the entire content, except within the hero blocks.

Rules:
* video links get transformed into vidyard player
* default (non-extra formatted) video links get transformed into vidyard player with lightbox
* italic video links get transformed into vidyard player embedded into the page
* video links (both styles) with link title other then the URL itself render the title blow the video as caption

PR #278, opened 2023-04-28
Simpler video

paolomoz:
Original:
https://www.moleculardevices.com/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/drafts/paolom/spectramax-i3x-readers1
- After: https://simpler-video--moleculardevices--hlxsites.hlx.page/drafts/paolom/spectramax-i3x-readers1


scripts/scripts.js
```diff
@@ -113,6 +113,42 @@ async function loadBreadcrumbs(main) {
...
+  let isV = false;
+  const hostnames = ['vids.moleculardevices.com', 'share.vidyard.com'];
+  [...hostnames].forEach((hostname) => {
+    if (url.hostname.includes(hostname)) {
+      isV = true;
+    }
+  });
+  return isV;
+}
+export function videoButton(container, button, url) {
```
mhaack, line 129: Do we foresee to use that somewhere else other than hero? If no I would put it into the hero-advanced.js

mhaack (approved):

scripts/scripts.js
paolomoz: this would allow to make any element clickable to open a video (button could be a simple link) not just the play icon of the hero. I see your point, if not used anywhere else don't even wonder. I will move to hero-advanced once I complete https://github.com/hlxsites/moleculardevices/issues/219 if it is actually never needed here. thanks

PR #288, opened 2023-05-03
product level RFQ changes

kapildhiman:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/quote-request?pid=spectramax-i-series-microplate-readers
- After: https://product-rfq--moleculardevices--hlxsites.hlx.live/quote-request?pid=spectramax-i-series-microplate-readers

Note: Not confirmed yet what will be the final value of Pid but to complete product level RFQ considering the pid is equal to the product URL. After getting confirmation from the business will make the final changes.


blocks/quote-request/quote-request.js
```diff
@@ -199,14 +228,21 @@ export default async function decorate(block) {
     htmlContentRoot.remove();
     const htmlContent = block.children[0].children[0].innerHTML.trim();
     block.innerHTML = `<div class="rfq-product-wrapper">
-    <div class="rfq-thankyou-msg">${htmlContent}</div>
-    </div>`;
+                       <div class="rfq-thankyou-msg">${htmlContent}</div>
```
lydiapuric, line 231: You are leveraging domHelpers already, and you are touching the file and change HTML. For readability please use the domHelpers here too to create the HTML.

blocks/quote-request/quote-request.js
```diff
@@ -199,14 +228,21 @@ export default async function decorate(block) {
...
+                       </div>`;
   } else {
+    const prfdData = await rfqData();
     parentSection.prepend(htmlContentRoot);
-    block.innerHTML = `
```
lydiapuric, line 206: Same here, please use domHelpers instead to improve readability.

lydiapuric (approved):

blocks/quote-request/quote-request.js
kapildhiman: @lydiapuric this one is also updated with domHelpers.

andreituicu (approved):
LGTM! 👍 

I don't see any option to improve the LHS -> TBT. We can't lazy load the iframe, because it is the main focal point of the page and it would not be experience for the user and the iframe is bringing a lot of 3rd party scripts.

PR #315, opened 2023-05-08
14 - Implement card / carousel combo block

### LINKED ISSUE #14: Implement card / carousel  combo block
This "standard" cards block is used on many overview pages.

- [ ] add hover effect on desktop
- [ ] change to carousel on mobile

Top page filters are via a separate issue #15 

Samples and variants:
* https://www.moleculardevices.com/applications > categorized view
* https://www.moleculardevices.com/technology > simple view
* https://www.moleculardevices.com/blog > simple view
* https://www.moleculardevices.com/products > categorized view, filtered by `landing-page-order` see below
* https://www.moleculardevices.com/products/additional-products > simple view, no filters
* https://www.moleculardevices.com/products/assay-kits > simple view, different card style
* https://www.moleculardevices.com/products/accessories-consumables > simple view, different card style
* https://www.moleculardevices.com/products/culture-media-reagents > simple view, different card style

Special handling for applications landing page
* use "card-descriptions" if present, fallback to "description"

Special filtering & sorting for product landing pages needed: 
* only display products with `landing-page-order` set to a number
* sort by number but within each product category, lowest number goes first

Special filtering for product sub category pages
* [Assay Kits](https://www.moleculardevices.com/products/assay-kits) > prefiltered to show assay kits only, use sub-category for filters
* [Accessories and Consumables](https://www.moleculardevices.com/products/accessories-consumables) > prefiltered by sub category "Accessories and Consumables" only, use main category for filters

Screenshot 2023-02-21 at 12 56 28

andreituicu:

Fix #14

**NOTE**: the cards in the lists are not pixel perfect like the ones from the original website, because we have unified the card implementation in all places with the `card.js`. This makes it more consistent across the whole website but leads to small inconsistencies with the original website.

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/applications
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/applications

- Before: https://main--moleculardevices--hlxsites.hlx.live/lab-notes/blog
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/lab-notes/blog

- Before: https://main--moleculardevices--hlxsites.hlx.live/products
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/products

- Before: https://main--moleculardevices--hlxsites.hlx.live/products/accessories-consumables
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/products/accessories-consumables

- Before: https://main--moleculardevices--hlxsites.hlx.live/products/assay-kits
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/products/assay-kits

- Before: https://main--moleculardevices--hlxsites.hlx.live/products/additional-products
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/products/additional-products

- Before: https://main--moleculardevices--hlxsites.hlx.live/technology
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/technology

- Before: https://main--moleculardevices--hlxsites.hlx.live/products/culture-media-reagents
- After: https://14-list-carousel--moleculardevices--hlxsites.hlx.live/products/culture-media-reagents


blocks/card/card.js
```diff
@@ -30,6 +31,8 @@ class Card {
   }
 
   renderItem(item) {
+    const cardTitle = item.h1 && item.h1 !== '0' ? item.h1 : item.title;
```
andreituicu, line 34: @duynguyen : I made the card title be the value of the `h1` if available. on the cards that I've tested it was closer to what the original site had, than the title. but I'm not fully sure if it is the same for all other cases. do you know by chance? do I need to add a config option for this?

paolomoz (approved):
@andreituicu looks good, a couple of notes

- (note for carousel swipe:) the original is more user friendly: when you swipe, the card moves already and when you release teh touch it scrolls to its position, while in Franklin version during swipe the card doesn't move and it only scrolls at the end, which makes the experience a bit uncomfortable
- it feels weird that we need 400+ lines of js for this on top of 500 lines of js for the carousel; I have not looked at the code deep enough to have a clear suggestion but high level: what's your opinion about so much code needed?
- what made you prefer having 2 different blocks for card filters and card list? does a card filter alone make sense? wondering why you are not including the filters as content of the card list block
- the look and feel of the cards is great :)

andreituicu:
@paolomoz Thank very much you for the review and the feedback! I'll answer to your questions in a different order, because the answers build on each other:

> what made you prefer having 2 different blocks for card filters and card list? does a card filter alone make sense? wondering why you are not including the filters as content of the card list block

We have at least 1 page (we'll see if more appear) where we have just the filter without the cards, like:
*  https://www.moleculardevices.com/lab-notes which is the same filter as in https://www.moleculardevices.com/lab-notes/blog
and vice versa, we have cards without filters, like:
* https://www.moleculardevices.com/products/additional-products

this is the reason why I went for 2 different blocks.

> it feels weird that we need 400+ lines of js for this on top of 500 lines of js for the carousel; I have not looked at the code deep enough to have a clear suggestion but high level: what's your opinion about so much code needed?

The reason why we need so much code, is because we have a lot of variations that we need to support. As example here are the very first implementations supporting a single usecase for
* card list (< 60 lines):   https://github.com/hlxsites/moleculardevices/blob/fd7123c54a42805bb2cb44098e750525d7e2e830/blocks/blog-list/blog-list.js

* carousel (~250 lines): https://github.com/hlxsites/moleculardevices/blob/e94aeb04710da450276446b33b83d80c6060e8f5/blocks/carousel/carousel.js

The complexity and code comes from the different variants that we need to support, for
* card list - with/without headings, with different types of sorting, clustering, data filtering, card rendering
* carousel - with/without infinite scrolling, with/without autoscrolling, with different renderings, with/without dot buttons, with/without counters, with different number of items in view for different screen sizes (e.g. 3 on desktop, 2 on tablet, 1 in mobile)

All these things add up and to give a comparison, the carousel library that they used has thousands of JS code lines with jQuery: https://github.com/OwlCarousel2/OwlCarousel2, same for its rewriting in plain JS: https://github.com/ganlanyuan/tiny-slider.

> (note for carousel swipe:) the original is more user friendly: when you swipe, the card moves already and when you release teh touch it scrolls to its position, while in Franklin version during swipe the card doesn't move and it only scrolls at the end, which makes the experience a bit uncomfortable

This is true indeed and comes from the fact that I opted for a very simplified implementation of the swipe behaviour. If the consensus is that the experience is not optimal, we can create an issue to implement a more complex swipe behaviour, but unfortunately that will come with even more code. 

I hope these answers are helpful! I'm happy to deep dive more if you have suggestions.

PR #321, opened 2023-05-08
fix: remove default h3 styles on product pages

duynguyen:
Fix #<gh-issue-id>

Test URLs:
- After: https://fix-product-h3--moleculardevices--hlxsites.hlx.page/products/clone-screening/mammalian-screening/cloneselect-imager

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -827,16 +827,10 @@ main .section.wave .default-content-wrapper p:first-of-type {
   margin-bottom: 10px;
 }
 
-.product .section.tabs h2,
-.section.tabs h3  {
+.product .section.tabs h2 {
   font-size: 36px;
   font-family: var(--ff-proxima-regular);
   margin-top: 0;
-}
-
-.product .section.tabs h3 {
-  font-size: 24px;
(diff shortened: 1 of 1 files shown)


duynguyen:
@lydiapuric - I need to revert the h3 styles on product pages because it breaks quite many places.

styles/styles.css
```diff
@@ -827,18 +827,12 @@ main .section.wave .default-content-wrapper p:first-of-type {
   margin-bottom: 10px;
 }
 
-.product .section.tabs h2,
-.section.tabs h3  {
```
lydiapuric, line 831: There are pages that have tabs but are not product pages, like the About Us page (/about-us)

mhaack: Hm, but should the style depend on the tabs? The it should be better per page theme or template if it is specific about the about is page.

duynguyen: @lydiapuric - are we good here to proceed?

lydiapuric:
@duynguyen Yes, go ahead.

lydiapuric (approved):

PR #353, opened 2023-05-14
IP Checker API integration to store user data

### LINKED ISSUE #182: Provide documentation and endpoint for the IP Checker API
We should call that API async to not delay the page loading. This can be done centrally via delay.js and the result should be stored in a cookie. That can be used by blocks and other logic to show region specific elements.

lydiapuric:

Fix #182

Test URLs:
- After: https://ipstack-integration--moleculardevices--hlxsites.hlx.page/


scripts/scripts.js
```diff
@@ -521,6 +511,44 @@ function setCookieFromQueryParameters(paramName, exdays) {
...
+      // eslint-disable-next-line no-console
+      console.warn('Could not load user information.', err);
+    });
+}
+
+/**
+ * loads everything that happens a lot later, without impacting
+ * the user experience.
+ */
+function loadDelayed() {
```
mhaack, line 545: Don’t touch `loadDelayed` here. The entire `loadUserData` function should go into `delayed.js`. I think we don’t even have to do anything for this in `scrips.js`.

mhaack (changes requested):

lydiapuric:
I wanted to use the functions for getting and setting the cookie from scripts.js. Putting te new function into delayed.js, and using functions from scripts.js, causes a circular dependency I wanted to avoid. On 15 May 2023, at 07:10, Markus Haack ***@***.***> wrote:﻿
@mhaack requested changes on this pull request.



In scripts/scripts.js:
> +      return Promise.reject(response);
+    }).then((data) => {
+      if (!getCookie(countryCodeInfo)) {
+        setCookie(countryCodeInfo, data[countryCodeInfo], 30);
+      }
+    }).catch((err) => {
+      // eslint-disable-next-line no-console
+      console.warn('Could not load user information.', err);
+    });
+}
+
+/**
+ * loads everything that happens a lot later, without impacting
+ * the user experience.
+ */
+function loadDelayed() {

Don’t touch loadDelayed here. The entire loadUserData function should go into delayed.js. I think we don’t even have to do anything for this in scrips.js.

scripts/scripts.js
lydiapuric: I see we ignored the circular dependency already  with drift. Moving... note: this is still a draft PR, the testing and update is missing

mhaack:
> I wanted to use the functions for getting and setting the cookie from scripts.js. Putting te new function into delayed.js, and using functions from scripts.js, causes a circular dependency I wanted to avoid. 

Understood, but I think this is fine. `Scrips.js` is always loaded before `delay.js` so nothing bad will happen. Also conceptuallly in Franklin everything loaded delayed should be in `delayed.js`.

lydiapuric:
ThanksYes, but still didn't like it ;) Added it now in delayed.js - however the PR is still WIP and still does not overcome CORS 🤪Soon, I will set up my localhost, so I can simulate the CORS issue locallyOn 15 May 2023, at 09:00, Markus Haack ***@***.***> wrote:﻿
Also check we already have // eslint-disable-next-line import/no-cycle from the boilerplate in delayed.js

scripts/scripts.js
```diff
@@ -521,6 +521,13 @@ function setCookieFromQueryParameters(paramName, exdays) {
   }
 }
 
+export function handleUpdateUserDataEvent(event, userData) {
```
mhaack, line 524: When is this function used?

lydiapuric: It is not used, I thought about adding a simple helper function for future events. I can remove that function.

lydiapuric: Function handleUpdateUserDataEvent is removed.

mhaack:
Code LGTM. Can not test as I run into `Error code 104 occured while loading user information. usage_limit_reached: Your monthly usage limit has been reached. Please upgrade your Subscription Plan.`

@kapildhiman I think we need the real account key here. The limit of the free key is too low and it will be blocked again quickly if we merge this to main.

jose-correia:
Code also LGTM.
Probably not related to these changes but the TBT looks a bit off-place here. Did you rebase with the newest changes to the megamenu, or is this related to another thing?

lydiapuric:
Thanks @mhaack and @jose-correia. @jose-correia I just merged with main, TBT looks good now.

mhaack (approved):

PR #361, opened 2023-05-15
fix TBT: delay newsletter iframe

andreituicu:

Fix #<gh-issue-id>

Test URLs:
- After: https://delay-newsletter--moleculardevices--hlxsites.hlx.live/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -173,8 +173,11 @@ export default async function decorate(block) {
   });
 
   buildNewsEvents(block.querySelector('.footer-news-events'));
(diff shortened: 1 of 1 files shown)


mhaack:
Was this only a test?

andreituicu:
@mhaack Initially I thought it was the proper fix I closed it because I didn't think it was good enough, as there is still some TBT and knowing I'm was starting the workshop. I wasn't super happy that it takes 3s to load the newsletter, so I was thinking  of a combination of `setTimeout` and `IntersectionObserver`, whichever happens first.

PR #364, opened 2023-05-16
chore: implement dynamic PID parameter for /quote-request URLs

mhaack:

Fix #<gh-issue-id>

Test URLs:
- After: https://<branch>--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -167,9 +167,10 @@ export function videoButton(container, button, url) {
   observer.observe(container);
 }
 
-function decorateEmbedLinks(main) {
+function decorateLinks(main) {
   main.querySelectorAll('a').forEach((link) => {
     const url = new URL(link.href);
+    // decorate video links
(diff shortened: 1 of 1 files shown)


lydiapuric (approved):
+1

blocks/header/helpers.js
```diff
@@ -73,3 +74,15 @@ export function reverseElementLinkTagRelation(element) {
 
   return element;
 }
+
+export function buildRequestQuote(classes) {
+  const familyId = getMetadata('family-id');
+  const link = familyId ? `/quote-request?pid=${familyId}` : '/quote-request';
```
lydiapuric, line 80: why not centralizing these 2 lines to build the search parameter for pid? you are using the same functionality in decorateLinks?

mhaack: In header it's special since `/quote-request` links is always built in code and was there before. I it already simplifying  a little as we not using the some code for desktop & mobile menu.

`decorateLinks` must cover more/all cases where we have other URL parameters as well or pid already set. This is not the case for the header, hence seperated.

PR #374, opened 2023-05-17
fix: load less data in footer

mhaack:

Fix #<gh-issue-id>

Test URLs:
- After: https://footer-json--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -53,6 +53,7 @@ function renderMoreLink(text, link) {
 async function renderEvents(container) {
   const events = await ffetch('/query-index.json')
     .sheet('events')
+    .chunks(5)
     .slice(0, 3)
     .all();
   container.innerHTML = '';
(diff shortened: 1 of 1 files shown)


blocks/featured-posts/featured-posts.js
```diff
@@ -4,6 +4,7 @@ import { createOptimizedPicture } from '../../scripts/lib-franklin.js';
 export default async function decorate(block) {
   const newsUrls = [...block.querySelectorAll('a')].map((a) => a.href);
   const newsItems = await ffetch('/query-index.json')
+    .chunks(500)
```
duynguyen, line 7: if these are news records only, isn't it better to fetch from the `news` sheet? less data to download and look up.

duynguyen: maybe not... just found out "featured posts" are `in-the-news` not `news` so they don't have a separate sheet.

duynguyen (approved):

PR #387, opened 2023-05-18
fix: Newsletter - High TBT

andreituicu:
Test URLs:
- After: https://newlettertbt--moleculardevices--hlxsites.hlx.live/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -105,8 +105,12 @@ function iframeResizeHandler(formUrl, id, container) {
 }
 
 async function buildNewsletter(container) {
-  loadScript('../../scripts/iframeResizer.min.js');
   const newsletterId = 'enewsletter';
+  if (container.querySelector(`#${newsletterId} iframe`)) {
+    return; // newsletter already present
+  }
+
+  loadScript('../../scripts/iframeResizer.min.js');
   const formId = 'enewsletterSubscribeForm';
   const formUrl = 'https://info.moleculardevices.com/newsletter-signup';
   const form = (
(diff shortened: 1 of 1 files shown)


blocks/footer/footer.js
```diff
@@ -173,8 +177,31 @@ export default async function decorate(block) {
...
+      observer.disconnect();
+      buildNewsletter(newsletterContainter);
+    }
+  });
+  observer.observe(newsletterContainter);
+
+  setTimeout(() => {
+    observer.disconnect();
+    buildNewsletter(newsletterContainter);
+  }, 3000);
```
lydiapuric, line 206: @andreituicu I understand your approach, but I don't understand what adding a delay of 3 sec will do for the user or the page. Why not stay with the Intersection Observer only and load the newsletter only on request?

andreituicu: @lydiapuric it makes it so that the newsletter is already loaded when the user reaches the footer (hopefully) in majority of cases and things don't move around. It's still CLS, even though it doesn't affect the LHS, because it is below the fold. I'm open to suggestions here 🙂 if that's too much of a detail to complicate the code so much, I'm ok to remove it.

mhaack: To avoid layout shift let’s add height/ width to the container of the iFrame so it already has the right size.

andreituicu: @mhaack Maybe I shouldn't have used the CLS term, I just meant that with the intersection observer the user can see the newsletter form not being present and then appearing after a moment, even if the container height would be set. But again, if this is too much detail, I can simply remove the `setTimeout` and leave just the `IntersectionObserver`.
My thinking was that it will be easier for me to delete it than to add it. 🙂

lydiapuric: @andreituicu Thanks for clarification. No, code is not too complex, I only think newsletter doesn't have to be loaded if the user never scrolls down to the page. 
However, IMO the newsletter subscription should be placed in a higher part of the site, but this is not of any concern here ;)

lydiapuric (approved):
LGTM +1

PR #449, opened 2023-05-26
Allow vidyard video in 'About Us' page to take full width

### LINKED ISSUE #421: [About us] Vidyard Video too small
Vidyard video on the first tab of the `About us` page is too small and opens a player (also too small) when clicked.

Original website
Screenshot 2023-05-24 at 13 43 01

Franklin
Screenshot 2023-05-24 at 13 42 35

lydiapuric:

Fix #421

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/about-us
- After: https://aboutus-vidyard--moleculardevices--hlxsites.hlx.live/about-us

### DIFF
--- a/templates/about-us/about-us.css
+++ b/templates/about-us/about-us.css
@@ -15,6 +15,10 @@
   font-size: 36px;
 }
 
+.about-us main .vidyard {
+  max-width: unset;
+}
+
 @media only screen and (max-width: 991px) {
   .about-us .section .default-content-wrapper h2 {
     font-size: 45px;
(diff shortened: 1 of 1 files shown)


templates/about-us/about-us.css
```diff
@@ -15,6 +15,10 @@
   font-size: 36px;
 }
 
+.about-us main .vidyard {
```
andreituicu, line 18: would it make sense to restrict it even more with `.section[data-name="about-us"]`? in case they add other videos in the other tabs they will not be affected?

lydiapuric:
@andreituicu .live gave even a worse LHS back than the  page, same range: around 60

PR #456, opened 2023-05-30
Landing Page (no header and special styling)

paolomoz:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/promotion/dd/img/slas-eu-2023
- After: https://landing-page-no-header--moleculardevices--hlxsites.hlx.page/en/assets/promotion/dd/img/slas-eu-2023


paolomoz:
closing with pending question whether this page should be actually ported, since it requires custom block/style and it refers to an event that was already held

PR #468, opened 2023-05-31
fix: temporary URL for sitemap.xml

mhaack:
Test URLs:
- After: https://sitemap-url--moleculardevices--hlxsites.hlx.live/


lydiapuric (approved):
+1

PR #494, opened 2023-06-01
fix columns alignment and margin top

paolomoz:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/applications/cell-counting/cell-counter-a431-cells
- After: https://cell-counter-a431-cells--moleculardevices--hlxsites.hlx.page/applications/cell-counting/cell-counter-a431-cells

### DIFF
--- a/blocks/columns/columns.css
+++ b/blocks/columns/columns.css
@@ -1,3 +1,7 @@
+main .section>div.columns-wrapper {
+  margin-top: 20px;
+}
+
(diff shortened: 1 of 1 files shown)


jose-correia (approved):

andreituicu:
Aren't these styles a bit too generic? Do they apply to all places where columns are used? Would it make sense to namespace it a bit?

PR #506, opened 2023-06-01
fix styling of en/assets pages

paolomoz:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/app-note/bpd/diverse-utility-of-cloneselect-imager-in-label-free-applications
- After: https://fix-en-assets--moleculardevices--hlxsites.hlx.page/en/assets/app-note/bpd/diverse-utility-of-cloneselect-imager-in-label-free-applications


styles/styles.css
```diff
@@ -227,7 +227,7 @@ h2, h3 {
 }
 
 h4 {
-  font-size: var(--heading-font-size-m)
+  font-size: var(--heading-font-size-l)
```
andreituicu, line 230: should this change rather be in a template or namespaced via a template class?

we would have to recheck if all the h4s match.

blocks/table/table.css
```diff
@@ -1,7 +1,8 @@
 .table {
   width: 100%;
   overflow-x: auto;
-  margin-top: 1em;
+  margin-top: 25px;
+  margin-bottom: 35px;
```
andreituicu, line 5: same here. should it be namespaced?

paolomoz: this is not an obvious call. if we assume we are almost perfect already, then we should make surgical changes. But my assessment is that we are not almost perfect. Hence even if there is risk of undesired overrides, on the other hand there could be the good effect of fixing all (or most) of the cases.
So instead of having a wrong general rule + many special rules, I am betting we improve the general rule.

Do you have any data points to help me make this call?

styles/styles.css
paolomoz: @andreituicu you are definitely right. What I was trying to solve her is more a content issue. so i put h4 back to size 'm'

The content issue is that in the original site there are headings in the content h3 and h4 with the same size; used alternatively and not sure if with any special criteria.
see for example "Measuring confluence in 96-well half-area plates" and "Conclusion" at:
https://www.moleculardevices.com/en/assets/app-note/bpd/diverse-utility-of-cloneselect-imager-in-label-free-applications

blocks/table/table.css
andreituicu: Unfortunately, I don't have more data points... I was trying to find the blog page where the table with caption is, but I can't find it, so I think it is good to go like this. A little bit more space can't hurt. 🙂

andreituicu (approved):

PR #537, opened 2023-06-02
css only caption styling

paolomoz:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/app-note/bpd/diverse-utility-of-cloneselect-imager-in-label-free-applications
- After: https://css-only-captions--moleculardevices--hlxsites.hlx.page/en/assets/app-note/bpd/diverse-utility-of-cloneselect-imager-in-label-free-applications
AND
- Before: https://main--moleculardevices--hlxsites.hlx.page/lab-notes/cellular-imaging-systems/3d-organoids-and-automation-of-complex-cell-assays
- After: https://css-only-captions--moleculardevices--hlxsites.hlx.page/lab-notes/cellular-imaging-systems/3d-organoids-and-automation-of-complex-cell-assays


andreituicu:
It doesn't work 100% for the list:
See https://css-only-captions--moleculardevices--hlxsites.hlx.page/lab-notes/clone-screening/how-to-develop-monoclonal-cell-lines-for-stem-cell-engineering at the end of the page, vs https://main--moleculardevices--hlxsites.hlx.page/lab-notes/clone-screening/how-to-develop-monoclonal-cell-lines-for-stem-cell-engineering

templates/application-note/application-note.js
```diff
@@ -1,7 +0,0 @@
-import { styleCaption } from '../../scripts/scripts.js';
```
mhaack, line 1: Why empty, we can remove the empty file

mhaack: There is CSS I know in application-note.css, but that can bey go to a block or styles.css

paolomoz: > Why empty, we can remove the empty file

I removed it now, anyway there is an empty file in the boilerplate too, hence took that as a reference 
https://github.com/adobe/helix-project-boilerplate/blob/main/blocks/hero/hero.js

andreituicu: @mhaack @paolomoz we need the js file and a `no-op` exported default function for template, otherwise we have errors in the console. Example: https://github.com/hlxsites/moleculardevices/blob/main/templates/about-us/about-us.js

 The alternative would be to improve the templating mechanism to support missing js or css files and not look for a default function in the js module.

andreituicu (approved):
LGTM! 👏  I should get into the habit of thinking more about selectors before resorting to JS. 🙂 

Please see my comment above about empty JS file for template https://github.com/hlxsites/moleculardevices/pull/537#discussion_r1214167577

andreituicu (approved):
LGTM! 👏  I should get into the habit of thinking more about selectors before resorting to JS. 🙂 

Please see my comment above about empty JS file for template https://github.com/hlxsites/moleculardevices/pull/537#discussion_r1214167577

templates/application-note/application-note.js
paolomoz: or we use Theme instead of Template; it makes styles.css larger but 1 less request for the imported css and 1 less js (so maybe 2 less requests). On the other hand that css is loaded for all pages and not only for the ones with that specific Theme

PR #546, opened 2023-06-02
fix: leadership page

### LINKED ISSUE #519: [Leadership] No breadcrumb
Where: https://main--moleculardevices--hlxsites.hlx.page/leadership

What: The page should have breadcrumb.

paolomoz:
Fix #519

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/leadership
- After: https://issue-519--moleculardevices--hlxsites.hlx.page/leadership

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -204,6 +204,7 @@ h1 {
   text-align: center;
 }
 
+header h1,
 .hero-inner h1 {
   text-align: unset;
 }
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -200,7 +200,12 @@ h6 {
 }
 
 h1 {
-  font-size: var(--heading-font-size-xxl)
+  font-size: var(--heading-font-size-xxl);
+  text-align: center;
+}
+
+.hero-inner h1 {
+  text-align: unset;
 }
 
 h2 {
@@ -236,14 +241,6 @@ h5 {
 
 h6 {
   font-size: var(--heading-font-size-xs)
-}
-
-h1 {
-  text-align: center;
-}
-
-.hero-inner h1 {
-  text-align: unset;
 }
 
 p,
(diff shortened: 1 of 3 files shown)


paolomoz:
The main fix here was to add /leadership breadcrumb:auto in metadata.xlsx

styles/styles.css
```diff
@@ -200,7 +200,12 @@ h6 {
 }
 
 h1 {
-  font-size: var(--heading-font-size-xxl)
+  font-size: var(--heading-font-size-xxl);
+  text-align: center;
```
jose-correia, line 204: This change affects other headers throughout the website. For instance in the header. Cant we make it more specific?

jose-correia (changes requested):

styles/styles.css
paolomoz: @jose-correia I made it general on purpose. The idea is: H1 is always centered if it is in the content and not if it is in the header. H1 cannot be anywhere, right? The header case is covered 2 lines below :) 
I see your point but at the same time I didn't want to fix this specific page only but potentially all occurrencies.
If you have an example where H1 (not in header) should be aligned different, it would be great so we can make this more specific

paolomoz:
I noticed that https://main--moleculardevices--hlxsites.hlx.page/leadership broke since yesterday, both in main and in this branch. I think this is due to a content change. 
I see this in console: 
`failed to load module for leader-profiles TypeError: Failed to fetch dynamically imported module: https://main--moleculardevices--hlxsites.hlx.page/blocks/leader-profiles/leader-profiles.js`

alexcarol:
> this page is not broken due to content change made by @alexcarol working on #551 If possible it would be better to work on draft content until the change is complete

Sorry about that, merged the fix, feel free to rebase.

jose-correia (approved):

PR #558, opened 2023-06-03
Use JSONP for cart interactions

### LINKED ISSUE #539: [Shopify] Fix the cart count
Shopify integration seems to work on backend using mode 'no-cors', however currently we are running into a CORS issue to read out the cart count because we retrieve an empty response from the cart endpoint. This causes the cart count not to be updated and to be 0.
Call the cart endpoint inside the add/cart endpoint to update the counters.

https://franklin.moleculardevices.com/products/assay-kits/igg-quantification-assays/valitatiter#ordering-options

andreituicu:

Fix #539 

**NOTE:** in order for the code to work, it needs to be pushed to the main branch. It will not work in PR state, because it needs the domain to be a subdomain of moleculardevices.com, otherwise cookies are not correctly passed and the cart doesn't work.

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/assay-kits/igg-quantification-assays/valitatiter#ordering-options
- After: https://jsonp-shopify--moleculardevices--hlxsites.hlx.page/products/assay-kits/igg-quantification-assays/valitatiter#ordering-options

### DIFF
--- a/blocks/ordering-options/ordering-options.js
+++ b/blocks/ordering-options/ordering-options.js
@@ -1,6 +1,6 @@
 import { detectStore, getCartItemCount, setCookie } from '../../scripts/scripts.js';
 import {
-  a, div, h3, i, p, span,
+  a, div, domEl, h3, i, p, span,
 } from '../../scripts/dom-helpers.js';
 
 const SHOP_BASE_URL = 'https://shop.moleculardevices.com';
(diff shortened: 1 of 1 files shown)


andreituicu:
One thing I couldn't figure out exactly, is when would https://github.com/hlxsites/moleculardevices/pull/558/files#diff-27a342799167911b4cb9b2959472ce6d25e3d13e02c39c2129c983a8ddd15abdR145-R155 be called. It didn't seem to be used for me, so I didn't change it.

lydiapuric (approved):
Code looks good to me, investigated in jsonp some time yesterday night. +1
We will need for testing shopify integration to merge. otherwise I see we still get a cors issue.

And this finetuning we can do as part of a next pr:
the cart is not initialized so we need to set and read the cart item count on loading the first time

andreituicu:
I did some cart initialisation here: 
https://github.com/hlxsites/moleculardevices/pull/558/files#diff-27a342799167911b4cb9b2959472ce6d25e3d13e02c39c2129c983a8ddd15abdR208-R209

initially, it will get the cart and set the counter cookie. not sure if this is enough.

lydiapuric:
Confirming that the initialisation works with your changes. Tested with franklin.moleculardevices.com

PR #605, opened 2023-06-09
[Issue 530] Fix filter behaviour to work with multiregion and multitype events

### LINKED ISSUE #530: [Events] Overview page region filter should not combine regions
Where: https://main--moleculardevices--hlxsites.hlx.page/events

What: Region filter currently only shows the regions from the existing events. If an event is both in NA and Europe, it should display in both filtered results of NA and Europe. There should not be a filter called "NA, Europe".

alexcarol:
Fix #530 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/archived-events
- After: https://issue-530--moleculardevices--hlxsites.hlx.page/archived-events

### DIFF
--- a/blocks/events/events.js
+++ b/blocks/events/events.js
@@ -7,10 +7,14 @@ import {
   div, input, label, span,
 } from '../../scripts/dom-helpers.js';
 
+function splitByComma(value) {
+  return value.split(',').map((s) => s.trim());
+}
+
 function prepareEntry(entry, showDescription, viewMoreText) {
-  entry.filterEventType = entry.eventType.split(', ')
+  entry.filterEventType = splitByComma(entry.eventType)
     .map(toClassName);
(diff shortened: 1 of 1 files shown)


blocks/events/events.js
```diff
@@ -8,8 +8,10 @@ import {
 } from '../../scripts/dom-helpers.js';
 
 function prepareEntry(entry, showDescription, viewMoreText) {
-  entry.filterEventType = toClassName(entry.eventType);
-  entry.filterEventRegion = toClassName(entry.eventRegion);
+  entry.filterEventType = entry.eventType.split(', ')
```
duynguyen, line 11: safer to split by ',' (no space) and trim the string, because sometimes space might be missed.

alexcarol: good idea, will do

lydiapuric (approved):
+1

PR #660, opened 2023-06-19
Render compare products based on specifications metadata

jose-correia:

Fix #<gh-issue-id>

Test URLs:
- After: https://400-render-compare--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/card/card.js
+++ b/blocks/card/card.js
@@ -105,7 +105,11 @@ class Card {
         c2aLinkBlock,
       ),
     );
-    if (item.productShowInFinder && item.productShowInFinder === 'Yes') {
+    if (
+      item.productShowInFinder
+      && item.productShowInFinder === 'Yes'
+      && item.specifications !== '0'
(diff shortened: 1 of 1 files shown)


blocks/card/card.js
```diff
@@ -105,7 +105,11 @@ class Card {
         c2aLinkBlock,
       ),
     );
-    if (item.productShowInFinder && item.productShowInFinder === 'Yes') {
+    if (
+      item.productShowInFinder
+      && item.productShowInFinder === 'Yes'
```
andreituicu, line 110: Do we still need to check  the `productShowInFinder` property? My understanding was that this only controls if the product appears in https://www.moleculardevices.com/product-finder

jose-correia: Makes sense that we drop this check. I don't think it should have nothing to do with the compare toggle

andreituicu (approved):

PR #668, opened 2023-06-20
feat: Locale Aware Blocks

andreituicu:
We should be able to test on https://dede-hlxsites-moleculardevices.onelink-translations.com/ only after the PR is merged.
An example can be found in the `ImageXpress Pico Automated Cell Imaging System
` section from: https://adobe.sharepoint.com/:w:/r/sites/HelixProjects/_layouts/15/Doc.aspx?sourcedoc=%7B96BC0074-5D6C-4277-AFE2-EE10C06AF536%7D&file=cellular-imaging-systems.docx&action=default&mobileredirect=true&cid=48319d58-4850-47d7-82b2-2200addbf2fb

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/video-gallery/cellular-imaging-systems
- After: https://locale--moleculardevices--hlxsites.hlx.live/video-gallery/cellular-imaging-systems

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -198,6 +198,17 @@ export async function fetchPlaceholders(prefix = 'default') {
  * @param {Element} block The block element
  */
 export function decorateBlock(block) {
+  const SUPPORTED_LOCALES = [
+    'en',
+    'jp',
+    'de',
+    'fr',
+    'ko',
+    'it',
+    'es',
+    'zh',
+  ];
+
   const shortBlockName = block.classList[0];
   if (shortBlockName) {
     block.classList.add('block');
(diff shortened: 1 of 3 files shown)


scripts/lib-franklin.js
```diff
@@ -207,6 +218,31 @@ export function decorateBlock(block) {
...
+    });
+
+    if (localesFound.length !== 0
+      && !localesFound.includes(document.documentElement.getAttribute('original-lang'))) {
+      // skip block decoration and remove content from blocks which are not displayed due to locale
+      block.setAttribute('data-block-status', 'loaded');
+      block.setAttribute('data-block-skip', true);
+      block.setAttribute('data-block-skip-reason', 'locale');
+      block.innerHTML = '';
+    }
```
jose-correia, line 245: The logic seems fine to me. I was a bit worried about the 2-character limit optimization, but after checking all the official locale codes, I don't think they will be adding anything that is >2 chars.

The only thing I would change is how much code we are adding to the decorate function. I would maybe add a separate method, and separate the concerns between `locale checking and element manipulation` to make the code more understandable

jose-correia (changes requested):

scripts/lib-franklin.js
andreituicu: @jose-correia : Thanks for the feedback! Does my last commit address it? I'd like to keep the `decorateBlock` to invoke the locale logic, so we don't have to call 2 functions in every place we call `decorateBlock`.

jose-correia: No no, this is exactly what I meant, just adding an additional call inside decorateBlock, like you did in the last commit 👍

jose-correia (approved):

mhaack (approved):

PR #669, opened 2023-06-20
added height to modal

### LINKED ISSUE #663: Leadership pop up is not opening in mobile

johan-t:
The problem was that the modal did not have an assigned height under 676px width. 

Fix #663 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/leadership
- After: https://issue-663--moleculardevices--hlxsites.hlx.page/leadership

### DIFF
--- a/blocks/leader-profiles/leader-profiles.css
+++ b/blocks/leader-profiles/leader-profiles.css
@@ -291,7 +291,9 @@
 }
 
 @media only screen and (max-width: 767px) {
-  .leader-profiles-modal {
+  .leader-profiles-modal-wrapper,
+  .leader-profiles-modal-body {
+    height: 500px;
     padding: 10px;
   }
 
--- a/blocks/leader-profiles/leader-profiles.css
+++ b/blocks/leader-profiles/leader-profiles.css
@@ -291,10 +291,16 @@
 }
 
 @media only screen and (max-width: 767px) {
-  .leader-profiles-modal-wrapper,
   .leader-profiles-modal-body {
-    height: 500px;
-    padding: 10px;
+    height: 80vh;
+    margin:10px;
+    bottom: 35px;
+  }
+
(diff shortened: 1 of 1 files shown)


blocks/leader-profiles/leader-profiles.css
```diff
@@ -291,7 +291,9 @@
 }
 
 @media only screen and (max-width: 767px) {
-  .leader-profiles-modal {
+  .leader-profiles-modal-wrapper,
```
jose-correia, line 294: There is a style class in line 307 that takes precedence over this one, so this hight is not applied to the wrapper

```css
.leader-profiles-modal-wrapper {
    height: 80vh;
    overflow-y: auto;
  }
```

blocks/leader-profiles/leader-profiles.css
```diff
@@ -291,7 +291,9 @@
...
+  .leader-profiles-modal-body {
+    height: 500px;
```
jose-correia, line 296: I think this height probably needs to be even bigger for smaller displays

I was testing with 600px for instance, and the modal should show the leader text description, currently, it can only show the image.

Current:
Screenshot 2023-06-20 at 10 38 41

Original website https://www.moleculardevices.com/leadership 
Screenshot 2023-06-20 at 10 38 37

jose-correia (changes requested):

blocks/leader-profiles/leader-profiles.css
johan-t: Yes you are right, I fixed it

blocks/leader-profiles/leader-profiles.css
johan-t: I improved it for small devices, right now it is not 1:1 to the original but I think the current version is an improvement to the original, especially for very small screens.

jose-correia (approved):

PR #703, opened 2023-06-23
display video in embed block

### LINKED ISSUE #698: [Media Gallery] Gallery Videos are not show
https://franklin.moleculardevices.com/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers

paolomoz:
Fix #698

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers
- After: https://issue-698--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers

### DIFF
--- a/blocks/embed/embed.css
+++ b/blocks/embed/embed.css
@@ -17,6 +17,10 @@ main .embed > div {
   justify-content: center;
 }
 
+main .embed.video > div {
+  display: block;
+}
+
 @media (min-width: 768px) {
   main .embed {
     width: 690px;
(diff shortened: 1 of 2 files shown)


paolomoz:
high TBT is not related to this change, see test executing after reverting this change entirely still has high TBT
https://github.com/hlxsites/moleculardevices/pull/703/commits/fc8c6191501497fa635e35d857acb3e1b0f060d0

mhaack:
Why do we need this? Why not jus remove the embed block?

paolomoz:
> Why do we need this? Why not jus remove the embed block?

I wrote and then deleted the same comment :) 
I was torn, but my answer would be: 2 options and authors can use what they think is easier for them (e.g. in the same fragment there are all embeds and the video should be different if embed is not supported for video.
From a code perspective it's about skipping things in the embed block, not duplicating code

mhaack (approved):

PR #705, opened 2023-06-23
Create custom columns type for lab notes

alexcarol:
Fix #490

Test URLs:
- Original: https://www.moleculardevices.com/lab-notes/cellular-imaging-systems/innovation-in-automated-high-content-imaging
- Before: https://main--moleculardevices--hlxsites.hlx.page/lab-notes/cellular-imaging-systems/innovation-in-automated-high-content-imaging
- After: https://issue-490--moleculardevices--hlxsites.hlx.page/lab-notes/cellular-imaging-systems/innovation-in-automated-high-content-imaging

PR #707, opened 2023-06-23
Ensure "customers we serve" columns are shown as columns on desktop

### LINKED ISSUE #499: [Products] Ceros frames are not displayed in Media Gallery
Ceros items are not displayed in Media Gallery.
Could be related to https://github.com/hlxsites/moleculardevices/issues/423

Example: https://main--moleculardevices--hlxsites.hlx.page/products/microplate-readers/absorbance-readers/spectramax-abs-plate-readers

Screenshot 2023-06-01 at 14 49 37

### LINKED ISSUE #699: [About us] incorrect 3 columns of collaboration in customers we serve tab
https://main--moleculardevices--hlxsites.hlx.page/about-us#customers-we-serve

The 3 items should grow horizontally, not vertically.

alexcarol:
Fix #699 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/about-us#customers-we-serve
- After: https://issue-499--moleculardevices--hlxsites.hlx.page/about-us#customers-we-serve

Notes: the content was also addressed, it was originally created as a 3x3 matrix where each row represented a column on the UI, rather than 3 columns.

### DIFF
--- a/blocks/columns/columns.css
+++ b/blocks/columns/columns.css
@@ -141,17 +141,6 @@ main .section>div.columns-wrapper {
   vertical-align: top;
 }
 
-.section .columns.collaboration > div picture,
-.section .columns.collaboration > div img {
-  display: block;
-  margin: auto;
-}
-
-.section .columns.collaboration > div img,
-.columns.layout-20-80.partners-list-columns ul li img {
-  width: auto;
-}
-
 .columns.layout-20-80.partners-list-columns ul li picture,
 .columns.layout-20-80.partners-list-columns ul li img {
   display: block;
@@ -183,7 +172,7 @@ main .section>div.columns-wrapper {
   justify-content: center;
 }
 
-.columns.columns-3-cols:not(.innovation) {
+.columns.columns-3-cols:not(.innovation,.collaboration) {
   margin-left: -15px;
   margin-right: -15px;
 }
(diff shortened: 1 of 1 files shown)


blocks/columns/columns.css
```diff
@@ -183,7 +172,7 @@ main .section>div.columns-wrapper {
   justify-content: center;
 }
 
-.columns.columns-3-cols:not(.innovation) {
+.columns.columns-3-cols:not(.innovation,.collaboration) {
```
mhaack, line 175: I know this was 'collaboration' before, but we should change it to a more generic and reusable name. WDYT?
Like 'center' as to me this is a 3 column with centered content.

This then can be maybe used for https://github.com/hlxsites/moleculardevices/pull/705 as well

alexcarol: It's a bit tricky, the only thing they have in common is title color and font-size, font-style (italic vs regular) and text-align are different (though I see now reason why text-align couldn't be justified everywhere). Do you think it's salvageable?

duynguyen: I also found this a bit specific to this place. I got a partially similar styling in https://github.com/hlxsites/moleculardevices/pull/728 for center small image but again different color, font size, etc.

PR #114, opened 2023-03-23
feat: implement in page tab navigation

mhaack:
Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#overview
- After: https://page-nav--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#overview

PR #771, opened 2023-06-29
Fix issue where image-gallery goes out of viewport when closing the lightbox

alexcarol:
Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/applications/3d-cell-models/organoids
- After: https://fix-gallery-drift--moleculardevices--hlxsites.hlx.page/applications/3d-cell-models/organoids

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -150,16 +150,6 @@
   box-sizing: border-box;
 }
 
-html {
-  scroll-behavior: smooth;
-}
-
-@media screen and (prefers-reduced-motion: reduce) {
-  html {
-    scroll-behavior: auto;
-  }
-}
-
 header {
   height: var(--nav-height);
 }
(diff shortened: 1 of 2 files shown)


styles/styles.css
```diff
@@ -150,16 +150,6 @@
   box-sizing: border-box;
 }
 
-html {
```
alexcarol, line 153: @mhaack this behaviour was superseding any input in "scrollIntoView", forcing the behaviour to always be "smooth". I had a look and it looks like it was introduced in https://github.com/hlxsites/moleculardevices/pull/114/files#diff-37f6471b306924bbb531d040e7ab55df5f83511246a411d22947462323516afaR107, do you have any context on why/if it was necessary?

mhaack: Ok, thanks understood

mhaack (approved):

PR #805, opened 2023-07-03
Fix German sticky navbar

jose-correia:

Fix #<gh-issue-id>

Test URLs:
- After: https://fix-german-sticky-navbar--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -1817,6 +1817,10 @@ header .actionable-card-submenu.customercare-portal > div:nth-child(2) > div:nth
     font-size: 14px;
   }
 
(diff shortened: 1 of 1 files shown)


blocks/header/header.css
```diff
@@ -1817,6 +1817,10 @@ header .actionable-card-submenu.customercare-portal > div:nth-child(2) > div:nth
     font-size: 14px;
   }
 
+  html[lang="en"] header .mainmenu-wrapper.sticky .menu-nav-category {
```
mhaack, line 1820: I think we should rather use the `OneLink..` classes for this. @andreituicu wdyt?

mhaack (approved):

Issue #430, opened 2023-05-24
Low LHS on product pages


duynguyen:
Product pages with lots of resources give bad LHS because they have to load lots of resources. We need to improve this.

andreituicu:
Unassigned myself as I'm going on PTO.
Current fixed in place are on the linked PRs:
* Lazy load vidyard scripts: https://github.com/hlxsites/moleculardevices/pull/447
* Lazy load blocks from hidden page tabs https://github.com/hlxsites/moleculardevices/pull/441

Current still known issues causing TBT:
* `header/helpers.js` CC @jose-correia ( this is spiking on any page, being in the header)
* `blocks/applications-carousel/applications-carousel.js` CC @duynguyen 

<img width="919" alt="Screenshot 2023-06-22 at 11 45 25" src="https://github.com/hlxsites/moleculardevices/assets/3651689/9ca98fac-4c17-416c-b945-10c173855550">

andreituicu:
there are more unattributable long tasks, which I'm not 100% sure where they are coming from, but what I've noticed is that commenting out the menu and applications carousel brings the TBT to a good level, so they might be connected, meaning fixing those could fix the Unattributables as well.

<img width="919" alt="Screenshot 2023-06-22 at 11 46 20" src="https://github.com/hlxsites/moleculardevices/assets/3651689/31c26e95-5d1d-46d7-b65a-218f8e4c1205">

andreituicu:
Keeping this one instead, because I has more historic context. Closing https://github.com/hlxsites/moleculardevices/issues/857 instead.

mhaack:
Closing as defered, can be re-opend if needed later.

PR #809, opened 2023-07-03
744 - [Product] Request Pricing CTA on hero should be agnostic to user region

### LINKED ISSUE #744: [Product] Request Pricing CTA on hero should be agnostic to user region
https://franklin.moleculardevices.com/products/flipr-penta-high-throughput-cellular-screening-system/screenworks-software

Request Pricing CTA should only be visible for US and EU regions. Currently it is visible for all the regions.

andreituicu:

Fix #744

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/products/flipr-penta-high-throughput-cellular-screening-system/screenworks-software
- After: https://fix-744--moleculardevices--hlxsites.hlx.live/products/flipr-penta-high-throughput-cellular-screening-system/screenworks-software

### DIFF
--- a/scripts/delayed.js
+++ b/scripts/delayed.js
@@ -1,7 +1,5 @@
 // eslint-disable-next-line import/no-cycle
 import { sampleRUM } from './lib-franklin.js';
-// eslint-disable-next-line import/no-cycle
-import { getCookie, setCookie } from './scripts.js';
 
 // Core Web Vitals RUM collection
 sampleRUM('cwv');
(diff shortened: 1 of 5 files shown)


scripts/scripts.js
```diff
@@ -669,6 +669,11 @@ export function detectStore() {
   return getCookie('country_code') === 'US';
 }
 
+export function detectPricingRequestAvailable() {
```
mhaack, line 672: This is only used in hero.js right, then lets keep it there.

mhaack (approved):

scripts/delayed.js
```diff
@@ -46,6 +47,11 @@ async function loadUserData() {
 
     if (response.ok) {
       const data = await response.json();
+      if (data[attrContinentCode]) {
+        setCookie(attrContinentCode, data[attrContinentCode], 1);
```
mhaack, line 51: After thinking about this a little I suggest we do some cleanup here:
a) use local storage instead of cookie
b) store JSON object in local storage containing both values
c) only use one event

scripts/scripts.js
andreituicu: I thought there might be more usecases. Sure, I'll move it.

scripts/delayed.js
andreituicu: @mhaack I'm a little bit reluctant to remove the usage of the existing cookie for the country code. I'm not sure if there aren't other scripts relying on it or logic that's immediately visible in our JS.
I can still do one single `geolocationUpdated` event. That one seems less impactful to me.

andreituicu: Done in [6ea090b](https://github.com/hlxsites/moleculardevices/pull/809/commits/6ea090b9fa681ce4763565ff02a1c65bf30416b7).

andreituicu:
can't seem to get 100 LHS due to https://github.com/hlxsites/moleculardevices/issues/430.

mhaack (approved):

PR #850, opened 2023-07-06
Issue 817 - Fix microplate readers styling

### LINKED ISSUE #817: Fix microplate-readers page issues
1. Centre align all the h2 titles and the text below it 
2. anchor links missing on the images below the intro para 
3. Update font size to be 34px for "Single-mode plate readers" and "Multi-mode plate readers" titles
4. under "Multi-mode plate readers" , remove the second image 
5. View all applications CTA is missing under "Applications and Research"

alexcarol:
Fix #817 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/microplate-readers
- After: https://issue-817--moleculardevices--hlxsites.hlx.page/products/microplate-readers


templates/category/category.css
mhaack, line 1: Please move this to styles css. Should work with `.category main h2` etc. as well. No need to add another file which adds a new request. Also this leads to a 404 as the logic try to load `category.js`.

mhaack (changes requested):

templates/category/category.css
alexcarol: Done, after applying the suggested changes and removing redundant css this PR becomes quite small (though I suspect it has to do with work that has happened in parallel to this PR)

mhaack:
LGTM. Did you cross check with other category pages?

mhaack (approved):

alinarublea (approved):

mhaack:
@alexcarol found one more detail on the Microplate Readers page, on mobile the color is wrong. Can you fix that as well please. 

IMG_7269

alexcarol:
> @alexcarol found one more detail on the Microplate Readers page, on mobile the color is wrong. Can you fix that as well please.
> 
> IMG_7269

I implemented a fix for that. The only drawback is that it changes the order of the "even" rows on mobile, but it dramatically simplifies CSS. If you think that could be a problem I can write an alternative implementation with the fixes, but it probably makes sense to do it in JS (and have a "light" and a "dark grey" class being applied depending on display size).

alexcarol:
> LGTM. Did you cross check with other category pages?

I did cross check with some and everything works as expected

styles/styles.css
```diff
@@ -1519,6 +1519,14 @@ main .section.background-molecules {
...
 }
 
+.category .default-content-wrapper h3 {
+  font-size: 34px;
+}
+
+.category .default-content-wrapper h3 a {
+  color: black;
+}
+
```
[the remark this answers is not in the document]
alexcarol: In the page linked in #862 it does cause some changes, but it does look closer to the original site style (compared to what we have now).

mhaack (approved):

PR #912, opened 2023-07-14
Fix clone screening issues

### LINKED ISSUE #860: Fix clone screening page issues
https://franklin.moleculardevices.com/products/clone-screening#qpix-400-series

- [x] On the customer breakthrough carousel, the images should be in a circle, globally
- [x] For the products section, the title should be h3: 40px and the subtitle h4: 30px.
Like this - 

image

- [x]  For the "Applications & Research" and "Related Products & Services" sections, left align the titles and text coming in the cards, globally.

alexcarol:
Fix #860 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/clone-screening
- After: https://issue-860--moleculardevices--hlxsites.hlx.page/products/clone-screening

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -1609,12 +1609,12 @@ main .section.background-molecules {
 }
 
 
-.product-blue-titles .section:not(.columns-2,:nth-child(2),:nth-last-child(1), :nth-last-child(2), :nth-last-child(3)) .default-content-wrapper > h2 {
+.section.product-blue-titles .default-content-wrapper > h2 {
   font-size: 40px;
   color: #008da9;
 }
 
-.product-blue-titles .section:not(.columns-2) .default-content-wrapper > h4 {
+.section.product-blue-titles .default-content-wrapper > h4 {
   font-size: 30px;
 }
 
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -1558,9 +1558,9 @@ main .section.background-molecules {
   mix-blend-mode: darken;
 }
 
-.title-center h2,
-.title-center h3,
-.title-center h4 {
+.section.title-center > .default-content-wrapper > h2,
+.section.title-center > .default-content-wrapper > h3,
+.section.title-center > .default-content-wrapper > h4 {
   text-align: center;
 }
 
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -1609,13 +1609,16 @@ main .section.background-molecules {
 }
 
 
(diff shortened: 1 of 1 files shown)


styles/styles.css
```diff
@@ -1531,9 +1531,17 @@ main .section.background-molecules {
...
-.title-center h4 {
+.section.title-h3-40 > .default-content-wrapper > h3 {
+  font-size: 40px;
+}
+
+.section.title-h4-30 > .default-content-wrapper > h4 {
+  font-size: 30px;
+}
+
+.section.title-center > .default-content-wrapper > h2,
```
alexcarol, line 1542: This is needed to prevent titles inside blocks from being affected by this.

styles/styles.css
```diff
@@ -1531,9 +1531,17 @@ main .section.background-molecules {
...
-.title-center h2,
-.title-center h3,
-.title-center h4 {
+.section.title-h3-40 > .default-content-wrapper > h3 {
+  font-size: 40px;
+}
+
+.section.title-h4-30 > .default-content-wrapper > h4 {
+  font-size: 30px;
+}
```
mhaack, line 1540: Is this again for a dedicated special styling or why is a new section style needed here. Then I would rather introduce a theme property and overwrite `--heading-font-size-l` & `--heading-font-size-m` for these theme.

I would argue this should be the same on all category pages? Then lets use the category template/themen for this.

alexcarol: Ok, so i've been trying to make sense of the styles, and I can't really find a proper pattern. In this page there's a few more titles that should  be changed, and it can be done with a theme, but it feels very fragile (I had to add exceptions for columns-2 and the first two and last three sections). The most straightforward way to implement this would probably be something like adding a style in the metadata of each section that needs it, maybe something like ".product-blue-titles" (based on the fact that in the original site such titles are also blue, which I assume should still be a requirement.

What do you think?

duynguyen: yeah IMO a theme style is probably too generic because this is too specific to the section you're updating.
`.product-blue-titles` doesn't sound bad to me in this context.

mhaack: Let's split this into 2 styles please, can be `blue-title` but it should only set it to blue. Not do extra font-size etc. This is confusing for the author

alexcarol: My idea would be to call it "product-titles" (which is title + subtitle) instead. Do you think that would be enough, or a separate "Blue title" would be necessary too?

mhaack (approved):

PR #919, opened 2023-07-17
chore: update stickyness behaviour of main menu

### LINKED ISSUE #915: Remove sticky from mega menu
Remove sticky behavior from the mega menu to allow content visibility beyond the menu when open
https://franklin.moleculardevices.com/

Steps to reproduce - 
1. Open the mega menu
2. Attempt to scroll down
3. Notice that the mega menu remains sticky, preventing further scrolling. Content beyond the mega menu is not visible.

Check the behavior on live - https://www.moleculardevices.com/

mhaack:
Fix #915 

Test URLs:
- After: https://menu-sticky--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -551,9 +551,11 @@ function enableStickyElements() {
     let stackedHeight = 0;
     STICKY_ELEMENTS.forEach((element, index) => {
       if (currentScrollPosition > offsets[index] - stackedHeight) {
-        element.classList.add('sticky');
-        element.style.top = `${stackedHeight}px`;
-        stackedHeight += element.offsetHeight;
+        if (element.classList.contains('mainmenu-wrapper') && !element.querySelector('[aria-expanded="true"]')) {
+          element.classList.add('sticky');
+          element.style.top = `${stackedHeight}px`;
+          stackedHeight += element.offsetHeight;
+        }
       } else {
         element.classList.remove('sticky');
         element.style.top = '';


scripts/scripts.js
```diff
@@ -551,9 +551,11 @@ function enableStickyElements() {
     let stackedHeight = 0;
     STICKY_ELEMENTS.forEach((element, index) => {
       if (currentScrollPosition > offsets[index] - stackedHeight) {
-        element.classList.add('sticky');
-        element.style.top = `${stackedHeight}px`;
-        stackedHeight += element.offsetHeight;
+        if (element.classList.contains('mainmenu-wrapper') && !element.querySelector('[aria-expanded="true"]')) {
```
duynguyen, line 554: AFAIK sticky elements are dynamic list. Beside the main menu as sticky element, there are other elements such as page tabs also sticky, and the change here disable their stickiness. IMO instead of setting stickiness for "only the non expanded main menu", we should set stickiness for all items in `STICKY_ELEMENTS` except "the main menu that is expanded".

mhaack: Hm, the more I think about it yes you are right. Let me see how this can be optimized.

mhaack: @duynguyen please check again, now checking globaly if some menu item is expanded or not.

duynguyen (approved):

scripts/scripts.js
duynguyen: LGTM thanks

PR #932, opened 2023-07-19
Prevent carousel from inheriting “center” from section

alexcarol:
Fix #912

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/clone-screening
- After: https://issue-912-centered-cards--moleculardevices--hlxsites.hlx.page/products/clone-screening

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -928,6 +928,10 @@ main .section.features-section .button-container {
   font-style: italic;
 }
 
+main .section.center > div:not(.carousel-wrapper) {
+  text-align: center;
+}
+
 /* Featured Asset section */
 main div.section.featured-asset {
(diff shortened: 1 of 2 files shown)


lydiapuric (approved):

styles/styles.css
```diff
@@ -928,6 +928,10 @@ main .section.features-section .button-container {
   font-style: italic;
 }
 
+main .section.center > div:not(.carousel-wrapper) {
```
duynguyen, line 931: can we keep the generic as is and make `main .section .carousel-wrapper` aligned left?
Initially I did this not thing but it started complicating things if you add more and more exceptions.

alexcarol: Done

duynguyen (approved):

PR #933, opened 2023-07-19
feat: use placeholders for hardcoded text

### LINKED ISSUE #914: [Geo Sites] Use placeholders for hardcoded text
... so that they can display translated text on geo sites.

duynguyen:
Fix #914 . This change is only visible after merged.

Test URLs:
- After: https://914--moleculardevices--hlxsites.hlx.page/

List of affected pages are listed below, to verify that they are not broken.


duynguyen:
I found out that some texts are translated even without a translation with the placeholders.
e.g. https://dede-hlxsites-moleculardevices.onelink-translations.com/products/cellular-imaging-systems/high-content-imaging/imagexpress-pico#resources -> look for "View All" to DE "Alle anzeigen".

blocks/product-finder/product-finder.js
```diff
@@ -76,7 +78,7 @@ function startOver(e) {
   currentTab.classList.remove(ACTIVE_CLASS);
 
   const titleEl = document.querySelector('.product-finder-wrapper .product-finder-tab-title');
-  titleEl.innerHTML = DEFAULT_TITLE;
+  titleEl.innerHTML = placeholders.selectProductType || DEFAULT_TITLE;
```
lydiapuric, line 81: This should be rather in the constant defined than here

blocks/product-finder/product-finder.js
```diff
@@ -288,8 +290,9 @@ async function stepOne(callback) {
 }
 
 export default async function decorate(block) {
+  placeholders = await fetchPlaceholders();
   block.prepend(
-    h3({ class: 'product-finder-tab-title' }, DEFAULT_TITLE),
+    h3({ class: 'product-finder-tab-title' }, placeholders.selectProductType || DEFAULT_TITLE),
```
lydiapuric, line 295: This should be rather in the constant defined than here

lydiapuric (approved):
Looks good to me.
Minor issue to improve: The placeholder in the productfinder should be rather in the constant defined than inside the code.

jose-correia (approved):

mhaack (approved):

blocks/product-finder/product-finder.js
duynguyen: Normally I would do the same. It's a bit tricky in this case, because the translated text is fetched dynamically and async in code, hence no way to set it as a constant.

blocks/product-finder/product-finder.js
duynguyen: Normally I would do the same. It's a bit tricky in this case, because the translated text is fetched dynamically and async in code, hence no way to set it as a constant.

Issue #895, opened 2023-07-12
[Applications] UI issues


MeghnaTewari:
1. Can we update the TOC section for some pages like it is on live? Grey background and bullets under the titles
https://franklin.moleculardevices.com/technology/luminescence
https://franklin.moleculardevices.com/technology/absorbance#overview
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/46b3f6b3-2501-4e79-9c17-37d5726cc9a9)

2. Fix "View ELISA protocol****Automate ELISA workflow" CTA
https://franklin.moleculardevices.com/applications/enzyme-linked-immunosorbent-assay-elisa

3. Center align "Latest Resources" globally.
 https://franklin.moleculardevices.com/technology/fluorescence 

4. Add space above the title and add line separator 
https://franklin.moleculardevices.com/technology/fluorescence-polarization
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/fab826ea-da00-41ac-8005-413b70ff155f)

Remove
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/48b5b1b1-6d7a-486f-9a7d-63449a98c10a)

5. Fix UI of the table and TOC bullets should be below the paragraph
https://franklin.moleculardevices.com/technology/time-resolved-fluorescence-trf-tr-fret-htrf 
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/31d39ca4-9abd-4ab2-83b8-9a88eab69340)
Screenshot from live
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/fab03de1-5a24-4031-8c98-c3e5bd575333)

6. Topic cards missing 
https://franklin.moleculardevices.com/technology/western-blot 

7.  Topic card images should be consistent globally with the dimensions - 555*282
 https://franklin.moleculardevices.com/applications/cell-counting
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/8fc1aa4f-5e67-433a-93ad-9862433c50ed)

8.  Add line separator 
https://franklin.moleculardevices.com/applications/cell-imaging/cell-painting
![image](https://github.com/hlxsites/moleculardevices/assets/138755144/454ec6d0-f8b3-4f8b-a095-c41a26dc4473)

9. Fix "See detailed workflow" CTA , Add space above all the titles and unbold caption under the video
https://franklin.moleculardevices.com/applications/live-cell-imaging 

[...]

PR #935, opened 2023-07-19
931 topic card

### LINKED ISSUE #931: [Cards] Align the topic card images in the same row based on title height
Align the topic card images in the same row based on title height, eg for Related Applications Cards.
This is a follow-up of #895 

Below you can see an example of a 3-liner title, so in this case, 2-line space is getting added to the left card and the space is getting managed.
https://www.moleculardevices.com/technology/absorbance

Screenshot 2023-07-18 at 18 39 42

lydiapuric:

Fix #931 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/technology/absorbance
- After: https://931-topic-card--moleculardevices--hlxsites.hlx.page/technology/absorbance

- Before: https://main--moleculardevices--hlxsites.hlx.page/technology/luminescence
- After: https://931-topic-card--moleculardevices--hlxsites.hlx.page/technology/luminescence

### DIFF
--- a/blocks/related-applications/related-applications.js
+++ b/blocks/related-applications/related-applications.js
@@ -11,10 +11,10 @@ async function renderFragment(fragment, block, className) {
 
 function alignTitles() {
   // eslint-disable-next-line consistent-return
-  const observer = new IntersectionObserver((entries) => {
+  const observer = new ResizeObserver((entries) => {
     entries.forEach((entry) => {
       const currHeights = [];
-      let alignedHeights = [];
+      const alignedHeights = [];
       const cards = entry.target.querySelectorAll('.related-app');
 
       // get current heights
@@ -27,11 +27,7 @@ function alignTitles() {
       });
 
       // calculate new heights
-      if (window.innerWidth < 992) {
-        // 1 col per row
-        // eslint-disable-next-line no-const-assign
-        alignedHeights = [...currHeights];
-      } else {
+      if (window.innerWidth > 991) {
         // 2 cols per row
         for (let i = 0; i < currHeights.length; i += 2) {
           const curr = currHeights[i];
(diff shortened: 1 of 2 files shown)


lydiapuric:
Meghna validated the fix in #895 and confirmed that it works as expected.

blocks/related-applications/related-applications.js
```diff
@@ -9,6 +9,60 @@ async function renderFragment(fragment, block, className) {
   block.append(fragment);
 }
 
+function alignTitles() {
+  // eslint-disable-next-line consistent-return
+  const observer = new ResizeObserver((entries) => {
+    entries.forEach((entry) => {
```
duynguyen, line 15: can we add a wrapper condition to all logics below, that only runs if the viewport width is min 992px? like:
```
if (window.matchMedia('only screen and (min-width: 992px)').matches) {
  entries.forEach...
}
```
it would then skip all the calculation on mobile mode (better for performance)

blocks/related-applications/related-applications.js
```diff
@@ -9,6 +9,60 @@ async function renderFragment(fragment, block, className) {
...
+      const currHeights = [];
+      const alignedHeights = [];
+      const cards = entry.target.querySelectorAll('.related-app');
+
+      // get current heights
+      cards.forEach((card, idx) => {
```
duynguyen, line 21: I was thinking instead of 3 loops, we could merge them into one loop. something like:
```
for (let i = 0; i < cards.length; i += 2) {
  const card1 = cards[i];
  const card2 = cards[i + 1];
  // get and set height as done here
}
```
also no need to check window width again here because it's done by the outer condition as mentioned above.
wdyt?

blocks/related-applications/related-applications.js
lydiapuric: @duynguyen not really. the point is if you resize from > 991 to <= 991, then this logic has to run to remove the set heights

blocks/related-applications/related-applications.js
lydiapuric: @duynguyen this might work, checking. the only part is that i have to remove in the first step all the set heights. but this should work outside the loop.

blocks/related-applications/related-applications.js
lydiapuric: @duynguyen and as far as I know I don't know the width of the window it was before resizing....

blocks/related-applications/related-applications.js
lydiapuric: simplified logic

duynguyen (approved):

PR #941, opened 2023-07-21
Issue 862 - Fix Cellular Imaging page issues

### LINKED ISSUE #862: Fix Cellular Imaging page issues
https://franklin.moleculardevices.com/products/cellular-imaging-systems

1. Centre align all the h2 titles
2. Add line separators below each product section
3. Remove the grey background showing below the carousel
4. For "Enhance your imaging systems" section place "TRANSMITTED LIGHT IMAGING" image and text in the second row and bold the text below all the images. Also, add the molecule background image for the section.
5.  Customization and Automation section should be in a grey background.

alexcarol:
Fix #862

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/cellular-imaging-systems
- After: https://issue-862--moleculardevices--hlxsites.hlx.page/products/cellular-imaging-systems


styles/styles.css
```diff
@@ -656,11 +656,20 @@ main .section.highlight {
 }
 
 main .section.background-grey,
-main .section.background-grey + .section.wave {
+main .section.background-grey + .section.background-grey.wave {
```
duynguyen, line 659: this will make all waves after a bg grey section start white above the wave (high contrast), e.g. https://issue-862--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers

Screenshot 2023-07-21 at 16 38 20

styles/styles.css
```diff
@@ -656,11 +656,20 @@ main .section.highlight {
...
   background-color: var(--grey-background-color);
   margin-top: 0;
 }
 
+main .section.background-grey-top-half::before {
```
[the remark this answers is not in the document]
duynguyen: also this makes some waves too aligned to the bottom.
Screenshot 2023-07-21 at 16 45 02

styles/styles.css
alexcarol: Your example actually has the same issue highlighted in #862, where we don't want the bottom part of the wave to be grey. It needs to be fixed using the style introduced here in the content. AFAIK this code was introduced for this page: https://issue-862--moleculardevices--hlxsites.hlx.page/leadership and I addressed the content there (by adding background-grey to the section with the wave).

styles/styles.css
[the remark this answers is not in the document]
alexcarol: What kind of testing do you have in mind? Can't we test it here directly?

blocks/columns/columns.css
```diff
@@ -206,6 +206,17 @@ main .section>div.columns-wrapper {
   margin-left: 0;
 }
 
+.columns.customization-automation h2 {
```
duynguyen, line 209: can we rename the style to something more reusable? e.g. "video-teaser"
also let's try to make the video full height (currently it's a bit shorter than the right column)

blocks/columns/columns.css
```diff
@@ -206,6 +206,17 @@ main .section>div.columns-wrapper {
...
+  font-size: 40px;
+}
+
+.columns.customization-automation > div > div:nth-child(even) {
```
duynguyen, line 213: instead of even child, we can make it a bit more generic like "all cells that does not contain a video-wrapper"

styles/styles.css
duynguyen: ack

styles/styles.css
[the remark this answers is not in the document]
duynguyen: I mean: as this affects many other pages, it's saver to create a separate PR and make sure we update all the affected blocks (content) before merge. Things like the "discovery wave" at the bottom are shared and easy to update, but the others like customer story carousel above are embedded to the content so we need to check one by one.

blocks/columns/columns.css
alexcarol: That would require js, as with css you can't really check for children's properties, do you think it's worth it? did you have anything in mind?

blocks/columns/columns.css
alexcarol: Done

blocks/columns/columns.css
duynguyen: yes could be similar to https://github.com/hlxsites/moleculardevices/blob/3a83b29c79a16d97e7e2ca55eaf8c11b84454b40/blocks/applications-carousel/applications-carousel.css#L12

styles/styles.css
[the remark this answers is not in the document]
alexcarol: I've modified the solution to be backwards compatible, let me know your thoughts.

blocks/columns/columns.css
alexcarol: Done

styles/styles.css
```diff
@@ -839,6 +855,14 @@ main .section.service-support.support-options h2 {
   }
 }
 
+.section.integrate-section {
```
alexcarol, line 858: @mhaack can we discuss this on monday?

duynguyen: @alexcarol I just saw Megna's comment that the gap is now too narrow. Therefore it makes sense to set one generic style `no padding bottom` to the upper section and leave the next section as is. wdyt?

alexcarol: Definitely makes sense now, will do

alexcarol: Fixed

duynguyen (approved):

PR #942, opened 2023-07-24
Remove duplicated selectors that were breaking the build

alexcarol:
Test URLs:
- After: https://fix-build--moleculardevices--hlxsites.hlx.page/


lydiapuric (approved):
Thanks for fixing

PR #953, opened 2023-07-25
Issue 943 - Fix CloneSelect Imager slides

### LINKED ISSUE #943: [Product] implement product carousel in clone select imager
Fix the "Streamlined Workflow" section as per the live site https://www.moleculardevices.com/products/clone-screening/mammalian-screening/cloneselect-imager

Centre align, update font size and the text  color 
image

Remove bold from the h3 titles 

This should be in a carousel instead of  separate slides

image
image

alexcarol:
Fix #943

Test URLs:
- After: https://issue-943--moleculardevices--hlxsites.hlx.page/products/clone-screening/mammalian-screening/cloneselect-imager


blocks/carousel/carousel.css
```diff
@@ -54,12 +54,19 @@ main .carousel-item-image img {
   max-height: 400px; /* default - same as .products */
 }
 
+main .block.carousel.orange-title h2 {
```
duynguyen, line 57: should we make it more generic to section title?
IIRC you added one before for `.section.blue-title`?

styles/styles.css
```diff
@@ -1784,3 +1817,35 @@ main .section.background-molecules {
...
+
+main .section.no-padding-bottom {
+  padding-bottom: 0;
+}
+
+main .section.no-padding-top {
+  padding-top: 0;
+}
+
+.section .block.main-title-bold h2 {
```
duynguyen, line 1842: can't we reuse the style above? `.section.main-title-bold > div.default-content-wrapper > h2`

styles/styles.css
```diff
@@ -1723,6 +1751,11 @@ main .section.background-molecules {
   color: #008da9;
 }
 
+.section.green-title .default-content-wrapper > h2,
```
duynguyen, line 1754: ok so you add the title color here. IMO we should have the same for orange and electric blue.

blocks/carousel/carousel.css
```diff
@@ -54,12 +54,19 @@ main .carousel-item-image img {
...
+  color: var(--orange-text-color);
```
duynguyen, line 58: minor thing: can we change this var to `--text-orange` to be consistent with other text colors?

alexcarol:
@duynguyen on title styles: I made the decision to apply title effects on sections *only* on the default content wrapper. This is to prevent their formats from being applied indiscriminately to blocks within the section. As in some cases I needed the titles to have specific colors within blocks, I've also added them to the blocks (using ".block.<style class>"). I can make the ones applied to only carousel to apply to all blocks, but I think we should further discuss if you don't think that distinction is useful, as it will mean adding more content changes (not just here, but on earlier blue-title content).

blocks/clone-screening-carousel/clone-screening-carousel.css
```diff
@@ -0,0 +1,65 @@
+.clone-screening-carousel .carousel-item {
```
duynguyen, line 1: implementation looks good, but can we rename this carousel to something more generic?
something like `product-workflow-carousel`

blocks/carousel/carousel.css
alexcarol: done

styles/styles.css
```diff
@@ -91,7 +93,6 @@
   --light-border-color-gray: #ddd;
   --dark-border-color-gray: #adb3b7;
   --orange-background-color: #ee7624;
-  --orange-text-color: #ee7624;
```
mhaack, line 94: This has changed from `--orange-text-color` to `--text-orange` is that double checked to not be used elsewhere?

alexcarol: Yes, it's not used anywhere else

blocks/clone-screening-carousel/clone-screening-carousel.css
duynguyen: yes I understand it's very specific. But we should still name things generic in code so that they can be reused in the future with minimal code update, without content update.
e.g. when people wants to add the same carousel to the "microplate readers" product.

alexcarol: I feel somewhat strongly against that sentiment. I chose this name because I feel this block *should not* be reused (and attempting to do so would cause more good than bad). If someone wants to extract the code - and has enough context to do so (used in 5 places?), I feel like it would be safer to create a new block and update the content to use the new block, rather than modify this block without being sure whether you might be breaking something. WDYT?

duynguyen: ok, thanks for the clarification. I'm pretty much convinced by the latest justification 👍 
let's run by @mhaack to make sure he agrees as well.

lydiapuric (approved):
LGTM +1

PR #1062, opened 2023-08-14
feat: Normalise Redirects URLs

andreituicu:
This PR reduces the need for need to create multiple redirect entries for URLs ending with `/` or having upper case characters.

**LIMITATION:** The feature only works if there is no actual page mapped to that URL, so Franklin serves the 404 page.

Example:
With only 1 redirect entry
Screenshot 2023-08-14 at 13 10 36

All of the following URLs will work:
https://handle-redirects--moleculardevices--hlxsites.hlx.page/test-redirects-case (via the xcel sheet)
https://handle-redirects--moleculardevices--hlxsites.hlx.page/test-redirects-case/
https://handle-redirects--moleculardevices--hlxsites.hlx.page/Test-Redirects-Case
https://handle-redirects--moleculardevices--hlxsites.hlx.page/Test-Redirects-Case/
https://handle-redirects--moleculardevices--hlxsites.hlx.page/TEST-REDIRECTS-CASE
https://handle-redirects--moleculardevices--hlxsites.hlx.page/TEST-REDIRECTS-CASE/

### DIFF
--- a/404.html
+++ b/404.html
@@ -3,6 +3,26 @@
   <head>
     <title>404: Page Not Found !</title>
     <script type="text/javascript">
+      const locationURL = new URL(window.location.href);
+      let urlAdjusted = false;
+      console.log(locationURL);
+      if (locationURL.pathname.endsWith('/')) {
+        locationURL.pathname = locationURL.pathname.substring(0, locationURL.pathname.length -1);
+        urlAdjusted = true;
+        console.log('here', locationURL);
+      }
+
+      if (locationURL.pathname.toLowerCase() !== locationURL.pathname) {
+        locationURL.pathname = locationURL.pathname.toLowerCase();
+        urlAdjusted = true;
+      }
+
+      if (urlAdjusted) {
+        window.location.href = locationURL.href;
+      }
+    </script>
+
+    <script type="text/javascript">
       window.isErrorPage = true;
       window.errorCode = '404';
     </script>


andreituicu:
Remove the helix bot comment as it is no relevant for this PR, we are redirecting to https://google.com for test purposes only.

lydiapuric:
The redirects we are handling here should be treated as permanent redirects (301) and I recommend not handling the redirects in the 404 because if the search bot gets a 404 a few times, it will deindex the page from the search results.

andreituicu:
> The redirects we are handling here should be treated as permanent redirects (301) and I recommend not handling the redirects in the 404 because if the search bot gets a 404 a few times, it will deindex the page from the search results.

My thinking was, any bot that accesses the `/test-redirects-case` redirect will still receive `301` and can index it. **This is URL should be used in any embeddings on pages.**

If I understood correctly from @dev-rajneeshkumar this is for users that accidentally type `/test-redirects-case/` or `/Test-Redirects-Case` to not receive the 404, but rather to go to the intended page.

lydiapuric:
Thanks @andreituicu I understand and agree in this case.

404.html
```diff
@@ -2,6 +2,26 @@
 <html>
   <head>
     <title>404: Page Not Found !</title>
+    <script type="text/javascript">
+      const locationURL = new URL(window.location.href);
+      let urlAdjusted = false;
+      console.log(locationURL);
```
lydiapuric, line 8: Please remove the console.log statement

dev-rajneeshkumar:
Yes @andreituicu expectation is same and also these urls are not mapped with any actual urls. We just got them as request to create a vanity url of an existing page.

404.html
```diff
@@ -2,6 +2,26 @@
...
+      if (locationURL.pathname.endsWith('/')) {
+        locationURL.pathname = locationURL.pathname.substring(0, locationURL.pathname.length -1);
+        urlAdjusted = true;
```
lydiapuric, line 11: when the pathname is not in the expected typing (lower case) the 404 redirecting will be called another time. why did you not add the toLowerCase to this part of code?

andreituicu: @lydiapuric I hope I understood the question correctly. The actual redirect from the 404 is only happening here at the end: https://github.com/hlxsites/moleculardevices/pull/1062/files#diff-128cc35baaf40bac8ae4d2932ca6e009311e8279f03f258e9cc83966f1d5b32eR21 . 

Setting the `urlAdjusted = true`, just helps to do exactly what you mention, not do a redirect multiple times.

lydiapuric: I understand this. But imagine we remove the / at the end, but the path still does not exist because it has upper case letters inside. This is where i think the 404 could be triggered twice

lydiapuric: Yes, I missed a part. You are completely right. Thanks

lydiapuric (approved):
+1

PR #1068, opened 2023-08-15
Issue 1031

### LINKED ISSUE #1031: Fix caption style and UI issues
https://www.moleculardevices.com/applications/synthetic-biology

Facing a couple of issues- 

1. Unable to update the caption font size to 14 px under this video.
image

2. This image size needs to be reduced so that there is no extra space above the 3rd bullet
image

Like this- 
image

3. Need to make view the video CTA outlined
image

4. Not able to reduce the extra space coming for certain sections

image
image

You can check the appropriate spacing for this page on this URL- https://1-md30.pantheonsite.io/applications/synthetic-biology

alinarublea:

Fix #1031 

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/applications/synthetic-biology
- After: https://1031--moleculardevices--hlxsites.hlx.page/applications/synthetic-biology


styles/styles.css
```diff
@@ -1635,6 +1647,14 @@ main .section.background-molecules {
   padding-top: 50px;
 }
 
+.application .section.tabs > div > p:has(div.video-wrapper) + p {
```
andreituicu, line 1650: `:has` is not supported by firefox.

alinarublea: I was checking and I don t know if :has can be replaced or if I can have something alternative in the css selector because the tree looks like.
```
<div class = "default-content-warapper"> 
  <p><div class="video-wrapper">
     <div class="video-container">.....
     </div>
  </p>
  <p>
   <em>.....
   </em>
  </p>
</div>
```

styles/styles.css
```diff
@@ -1633,7 +1633,7 @@ main .section.background-molecules {
   padding-top: 0
 }
 
-#what-are-the-benefits-of-synthetic-biology {
+#what-are-the-benefits-of-synthetic-biology, #what-are-some-synthetic-biology-applications, #cellular-agriculture, #cultured-meat {
```
lydiapuric, line 1636: I would rather see to introduce an option in the content instead using the ids to address this spacing. WDYT?

styles/styles.css
lydiapuric: Unfortunately there is no replacement. In worst case you have to take care to achieve the same as a code change.

PR #1101, opened 2023-08-22
fix: Set campaing ID from cookie of present for get in touch

### LINKED ISSUE #1103: Set campaing ID from cookie if present for RFQ, landing pages and get in touch
If user lands on website with some cmp id value in the query string like https://www.moleculardevices.com/events?cmp=1234, then cookie will be set for that particular cmp id and leads will be assigned to that particular cmp id which is there in the cookie/query string  rather than default cmp id for that particular page/form.

andreituicu:

Fix #1103

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/contact?cmp=test
- After: https://getintouch-cmp--moleculardevices--hlxsites.hlx.live/contact?cmp=test

### DIFF
--- a/blocks/get-in-touch/get-in-touch.js
+++ b/blocks/get-in-touch/get-in-touch.js
@@ -1,4 +1,4 @@
-import { getCookie } from "../../scripts/scripts.js";
+import { getCookie } from '../../scripts/scripts.js';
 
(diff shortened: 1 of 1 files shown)


alinarublea (approved):

mhaack (approved):
LGTM but please fix the lint issue

PR #1112, opened 2023-08-23
feat: Regenerate the Coveo xml when a new path is published

### LINKED ISSUE #1099: Regenerate the Coveo xml when a new path is published/removed
Currently updating the coveo xml is a manual task:
```
git checkout main
cd coveo
node generate-coveo-sitemap.js
git add coveo-xml.xml
git commit
git push
```

We should automate this procedure to be executed everytime a page is published/removed using a github action and the Franklin github hook.

See https://github.com/adobe/helix-website/blob/main/.github/workflows/log-publish.yml as an example github action.

The action should be triggered only when `/query-index.json` is published

duynguyen:

Fix #1099

Test URLs:
- After: https://1099-coveo--moleculardevices--hlxsites.hlx.page/

### DIFF
--- a/.github/workflows/generate-coveo-index.yaml
+++ b/.github/workflows/generate-coveo-index.yaml
@@ -7,20 +7,17 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
+      - name: Regenerate coveo.xml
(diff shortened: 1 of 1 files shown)


duynguyen:
Checked the script execution in https://github.com/hlxsites/moleculardevices/actions/runs/5952543573.
Custom event `repository_dispatch -> resource-published` can only be tested once code is merged to main

PR #1129, opened 2023-08-25
feat: direct access to gated resources if user is authorized

### LINKED ISSUE #1104: Give access to gated resources if user is logged in https://support.moleculardevices.com/
If user is already logged in the support portal i.e. https://support.moleculardevices.com/ then a cookie will be created and on the basis of that value in moleculardevices.com, all gated resources will appear as ungated to that particular user.

Example cookie value: 
image

Here are the all possible values of cookie

DISTRIBUTOR
INTEGRATOR
SALES
TECH

**Regards,
Rajneesh**

duynguyen:

Fix #1104 
To be an authorized user, set cookie `STYXKEY_PortalUserRole` to one of these values:
DISTRIBUTOR
INTEGRATOR
SALES
TECH

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#resources
- After: https://1104--moleculardevices--hlxsites.hlx.live/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#resources

### DIFF
--- a/blocks/card/card.js
+++ b/blocks/card/card.js
@@ -5,7 +5,7 @@
 import {
   decorateIcons, loadCSS, createOptimizedPicture, fetchPlaceholders, toCamelCase,
 } from '../../scripts/lib-franklin.js';
-import { summariseDescription } from '../../scripts/scripts.js';
+import { isGatedResource, summariseDescription } from '../../scripts/scripts.js';
 import {
   a, div, h3, p, i, span,
 } from '../../scripts/dom-helpers.js';
@@ -84,7 +84,7 @@ class Card {
       ? item.imageBlock : createOptimizedPicture(itemImage, item.title, 'lazy', [{ width: '800' }]);
 
     let cardLink = item.path;
-    if (item.gated === 'Yes' && item.gatedURL && item.gatedURL !== '0') {
+    if (isGatedResource(item) && item.gatedURL && item.gatedURL !== '0') {
       cardLink = item.gatedURL;
     } else if (item.redirectPath && item.redirectPath !== '0') {
       cardLink = item.redirectPath;
--- a/blocks/card/card.js
+++ b/blocks/card/card.js
@@ -84,7 +84,7 @@ class Card {
       ? item.imageBlock : createOptimizedPicture(itemImage, item.title, 'lazy', [{ width: '800' }]);
 
     let cardLink = item.path;
-    if (isGatedResource(item) && item.gatedURL && item.gatedURL !== '0') {
+    if (isGatedResource(item)) {
       cardLink = item.gatedURL;
     } else if (item.redirectPath && item.redirectPath !== '0') {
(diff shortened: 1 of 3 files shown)


blocks/card/card.js
```diff
@@ -84,7 +84,7 @@ class Card {
       ? item.imageBlock : createOptimizedPicture(itemImage, item.title, 'lazy', [{ width: '800' }]);
 
     let cardLink = item.path;
-    if (item.gated === 'Yes' && item.gatedURL && item.gatedURL !== '0') {
+    if (isGatedResource(item) && item.gatedURL && item.gatedURL !== '0') {
```
alexcarol, line 87: Why doesn't this function include `item.gatedURL && item.gatedURL !== '0'`? It looks like that's a requirement for resources to be considered gated too.

duynguyen: good point, thank you! fixed.

lydiapuric:
Hmm, tried it on https://1104--moleculardevices--hlxsites.hlx.live/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#resources, cleared all cookies, not setting of STYXKEY_PortalUserRole cookie
I still can read the full application note...

duynguyen:
@lydiapuric did you see some app notes that are gated but having direct link on this page? can you share the app note name so I can check?

lydiapuric:
@duynguyen NanoBRET: https://1104--moleculardevices--hlxsites.hlx.live/en/assets/app-note/br/nanobret-technology-on-spectramax-i3x-multi-mode-microplate-readers

lydiapuric:
NanoBRET is not gated, I will recheck with a gated one. Sorry for the noise :(

lydiapuric:
@duynguyen , when I know the gated url, I can access that page without having the cookie checked. IMO we need to prohibit this resource in that case too

duynguyen:
@lydiapuric yes that's correct - anyone knowing the file URL can access it directly without being "authorized" (which is soft anyway as you can simply set the cookie client side).

I'm not sure if we should do it differently though, as that's how the customer wants it in the old site as well. tagging @dev-rajneeshkumar to confirm.

mhaack:
@lydiapuric @duynguyen I think this was the same before. The entire gated url concept is soft only. If you know the doc url you can always bypass the landing page.

lydiapuric:
@duynguyen No, didn't find a gated App Note

lydiapuric (approved):
LGTM +1
Soft access is addressed and works.

lydiapuric:
@mhaack and @duynguyen Just checked https://live-md30.pantheonsite.io/sites/default/files/resources/tagging/elisa.pdf and you are right, the gated resource is also here accessible with the direct URL.

alexcarol (approved):

mhaack (approved):

PR #1149, opened 2023-08-30
Issue 1143 - Redirect authorized users automatically in gated pages

### LINKED ISSUE #1143: Gated resource links within content should be ungated when cookie is set
If a link to the gated resource is present within the content and a cookie is set, the resource should be ungated for that user. 

Example - https://1104--moleculardevices--hlxsites.hlx.page/lab-notes/cellular-imaging-systems/case-study-assay-ready-patient-derived-organoids-and-high-throughput-3d-imaging
image

alexcarol:
Fix #1143

How to test: in the console use document.cookie='STYXKEY_PortalUserRole=DISTRIBUTOR'; and refresh the page. If the cookie is present the page should redirect to the new page.

Test URLs:
Regular gated to ungated page:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/app-note/br/optimizing-performance-of-transcreener-fluorescence-polarization-assays-with-the-spectramax-id5-multi-mode-microplate-reader
- After: https://issue-1143--moleculardevices--hlxsites.hlx.page/en/assets/app-note/br/optimizing-performance-of-transcreener-fluorescence-polarization-assays-with-the-spectramax-id5-multi-mode-microplate-reader

Gated to pdf:
- Before: https://main--moleculardevices--hlxsites.hlx.page/en/assets/ebook/dd/img/essential-guide-to-organoids-in-drug-discovery
- After: https://issue-1143--moleculardevices--hlxsites.hlx.page/en/assets/ebook/dd/img/essential-guide-to-organoids-in-drug-discovery

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -1070,12 +1070,16 @@ cookieParams.forEach((param) => {
   setCookieFromQueryParameters(param, 0);
 });
 
+export function isAuthorizedUser() {
+  const supportCookie = getCookie('STYXKEY_PortalUserRole');
+  return supportCookie && SUPPORT_CHANNELS.includes(supportCookie);
+}
+
 /**
  * Check if a resource should be served as gated or original
  */
 export function isGatedResource(item) {
-  const supportCookie = getCookie('STYXKEY_PortalUserRole');
-  const authorizedUser = supportCookie && SUPPORT_CHANNELS.includes(supportCookie) && window.location.hostname.endsWith('moleculardevices.com');
+  const authorizedUser = isAuthorizedUser() && window.location.hostname.endsWith('moleculardevices.com');
   return item.gated === 'Yes' && item.gatedURL && item.gatedURL !== '0'
     && (!authorizedUser || item.gatedURL.includes('https://chat.moleculardevices.com'));
 }
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -1072,14 +1072,14 @@ cookieParams.forEach((param) => {
 
 export function isAuthorizedUser() {
   const supportCookie = getCookie('STYXKEY_PortalUserRole');
-  return supportCookie && SUPPORT_CHANNELS.includes(supportCookie);
+  return supportCookie && SUPPORT_CHANNELS.includes(supportCookie) && window.location.hostname.endsWith('moleculardevices.com');
 }
 
 /**
  * Check if a resource should be served as gated or original
  */
 export function isGatedResource(item) {
-  const authorizedUser = isAuthorizedUser() && window.location.hostname.endsWith('moleculardevices.com');
+  const authorizedUser = isAuthorizedUser();
   return item.gated === 'Yes' && item.gatedURL && item.gatedURL !== '0'
     && (!authorizedUser || item.gatedURL.includes('https://chat.moleculardevices.com'));
 }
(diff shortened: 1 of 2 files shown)


scripts/scripts.js
```diff
@@ -1070,12 +1070,16 @@ cookieParams.forEach((param) => {
   setCookieFromQueryParameters(param, 0);
 });
 
+export function isAuthorizedUser() {
+  const supportCookie = getCookie('STYXKEY_PortalUserRole');
+  return supportCookie && SUPPORT_CHANNELS.includes(supportCookie);
+}
```
mhaack, line 1076: I think we have some helper for that already @duynguyen ?

templates/landing-page/landing-page.js
```diff
@@ -12,7 +13,18 @@ async function iframeResizeHandler() {
...
 }
 
-function handleEmbed() {
+async function handleEmbed() {
+  if (isAuthorizedUser()) {
+    const path = window.location.pathname;
+    const pageIndex = await ffetch('/query-index.json').all();
+    const foundPage = pageIndex.find((page) => page.gated === 'Yes'
+      && page.gatedURL
+      && page.gatedURL.includes(`moleculardevices.com${path}`));
```
duynguyen, line 22: could `endWith` be better here to look for the exact page. `includes` might also retrieve the child pages under the same path (if any)

templates/landing-page/landing-page.js
```diff
@@ -12,7 +13,18 @@ async function iframeResizeHandler() {
   iFrameResize({ log: false });
 }
 
-function handleEmbed() {
+async function handleEmbed() {
```
duynguyen, line 16: now that the function is async, the caller where it is called should also async - await.

scripts/scripts.js
duynguyen: this is good as Alex decoupled the `isAuthorized` logics from the helper before, so that it can be reused to check the condition before fetching the query index for unauthorized users.
@alexcarol one possible enhancement in this function is to move the `window.location.hostname.endsWith('moleculardevices.com')` logic to here, because the "authorized user" concept only applies on this domain (as requested by @dev-rajneeshkumar).

templates/landing-page/landing-page.js
alexcarol: done

templates/landing-page/landing-page.js
alexcarol: Done

templates/landing-page/landing-page.js
```diff
@@ -12,7 +13,16 @@ async function iframeResizeHandler() {
...
+  if (isAuthorizedUser()) {
```
mhaack, line 17: I wonder why we get this into the handleEmbed() at all. Can't we just put the check into buildAutoBlocks directly after the thank you page check?

alexcarol: Do we even want it after the thank you page check? I'd assume we want to redirect any time they have the cookie (if the cookie is preset they'll never see the gated page anyway).

scripts/scripts.js
alexcarol: As discussed in the call I'll move the logic to isAuthorized, but would appreciate an "approve" before doing so to allow testing in the branch should it be needed.

templates/landing-page/landing-page.js
alexcarol: Change made, lmk if you're fine with having it before "thank you"

mhaack (approved):

duynguyen (approved):

PR #1160, opened 2023-09-01
Fix card width in product finder for desktop to address translations

### LINKED ISSUE #1113: The product finder is using shared data for front and backend operations
The product finder is still using shared data for front and backend operations. Ideally, this should be split. Categories in step 2 cannot be translated without breaking functionality.

On step three no products are displayed. This appears to be caused by another mismatch between translated and untranslated data.

In /blocks/product-finder/product-finder.js

async function stepThree(e) {
…
let category = getTabTitle(e.target);
const originalCategory = category;
…
const categoryData = categories.find(
  (c) => c.category === originalCategory && c.type === originalType,
);

The “category” value retrieved from the “getTabTitle” function is translated. Later on this value is compared to categories directly from the json, which is untranslated. In order for the last step to work the tab title (data-title attribute) should be in English while the displayed text (span.icon-title) is translated.

**Note: This is an issue pointed out by Transperfect and needs to be fixed by Adobe Team.**

alexcarol:
Fix #1113

Test URLs:
How to test: select Instruments > Cellular Imaging systems
- Before: https://main--moleculardevices--hlxsites.hlx.page/product-finder
- After: https://issue-1113--moleculardevices--hlxsites.hlx.page/product-finder
- Before german: https://dede-hlxsites-moleculardevices.onelink-translations.com/product-finder
- After german: https://issue-1113-de--moleculardevices--hlxsites.hlx.page/product-finder

### DIFF
--- a/blocks/product-finder/product-finder.css
+++ b/blocks/product-finder/product-finder.css
@@ -441,10 +441,6 @@
     width: 300px;
   }
 
-  .product-finder-container #step-3 .card-caption .button-container {
-    margin-bottom: 0;
-  }
-
   .product-finder-container #step-3 .card-caption .compare-button {
     margin-top: 7px;
   }
--- a/blocks/product-finder/product-finder.css
+++ b/blocks/product-finder/product-finder.css
@@ -436,4 +436,16 @@
     width: 33.35%;
     border-left: unset;
   }
+
+  .product-finder-container #step-3 .card {
+    width: 300px;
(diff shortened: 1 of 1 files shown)


jose-correia:
Screenshot 2023-09-01 at 18 18 11

My only input is that i think this padding-bottom should be a bit bigger

jose-correia:
Will approve already

PR #1180, opened 2023-09-11
fb posts

dev-rajneeshkumar:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/lab-notes
- After: https://socialimedia--moleculardevices--hlxsites.hlx.page/lab-notes

### DIFF
--- a/blocks/embed/embed.css
+++ b/blocks/embed/embed.css
@@ -29,6 +29,10 @@ main .embed > div {
   justify-content: center;
 }
 
+main .embed.embed-facebook > div{
+  display: block;
+}
+
 main .embed.video > div {
   display: block;
 }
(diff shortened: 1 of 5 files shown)


mhaack (commented):
While this works, I think this should be better done via the embed block we already have. This can already handle multiple externe embeds and the Facebook feed can be added here.

dev-rajneeshkumar:
> While this works, I think this should be better done via the embed block we already have. This can already handle multiple externe embeds and the Facebook feed can be added here.

Thanks for your comments @mhaack , we have updated the code via embed. Please review and approve.

mhaack (approved):

PR #1193, opened 2023-09-13
feat: disable delay.js and footer loading for /cp-request call

### LINKED ISSUE #1187: GA & GTM - filter cp-request traffic
Request from Business:

Since migrating to Adobe we've seen a high volume of traffic going to the pages that contain "cp-request". Attached is the GA page report showing pageviews to URLs containing "cp-request" have increased starting in August. Please filter out these pages from our GA so the traffic generated is not shown so we don't mislead our reporting. 
Secondly, we need to look in GTM and add a trigger to exclude the GA script from firing on these pages so we don't experience future traffic. 
Please confirm once completed. 
[Analytics Molecular Devices Filtered View Minus Softmax Pro Pages 20230701-20230911.pdf](https://github.com/hlxsites/moleculardevices/files/12594666/Analytics.Molecular.Devices.Filtered.View.Minus.Softmax.Pro.Pages.20230701-20230911.pdf)

mhaack:
Fix #1187

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/cp-request
- After: https://cp-request--moleculardevices--hlxsites.hlx.page/cp-request?STYXKEY_PortalUserRole=cplogout

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -952,26 +952,28 @@ async function loadLazy(doc) {
 
   await loadBlocks(main);
 
-  enableStickyElements();
+  if (!window.location.pathname.startsWith('/cp-request')) {
+    enableStickyElements();
 
-  const { hash } = window.location;
-  const element = hash ? doc.getElementById(hash.substring(1)) : false;
-  if (hash && element) element.scrollIntoView();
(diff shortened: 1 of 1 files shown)


lydiapuric (approved):
+1

PR #1262, opened 2023-10-27
use .target to populate rum

MarquiseRosier:
This ensures rum data passes through

https://moleculardevices.com


mhaack:
@MarquiseRosier we also have `rfqRUM.type = hubSpotQuery.requested_qdc_discussion__c;` would we need to adjust that as well?

mhaack:
The thinking behind ist the RFQ is either a "global" or "product" RFQ, this is what we put into `rfqRUM.source`. If product we have the product identifier (aka. familyID) as well to know which product. But also the RFQ has a "type" which can be "Call", "Demo", ... can we get this attached to RUM as well?

mhaack (approved):

PR #1287, opened 2024-01-02
latest resources carousel sorting

dev-rajneeshkumar:
To  add date under resource type in latest resources carousel 

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/technology/ai-machine-deep-learning-cell-image-analysis#overview
- After: https://latest-resources--moleculardevices--hlxsites.hlx.page/technology/ai-machine-deep-learning-cell-image-analysis#overview

### DIFF
--- a/blocks/card/card.js
+++ b/blocks/card/card.js
@@ -62,7 +62,7 @@ class Card {
     this.c2aLinkStyle = false;
     this.c2aLinkConfig = false;
     this.c2aLinkIconFull = false;
-    this.showDate = true,
+    this.showDate = true;
 
     // Apply overwrites
     Object.assign(this, config);
--- a/blocks/card/card.js
+++ b/blocks/card/card.js
@@ -5,7 +5,7 @@
 import {
   decorateIcons, loadCSS, createOptimizedPicture, fetchPlaceholders, toCamelCase,
 } from '../../scripts/lib-franklin.js';
-import { isGatedResource, summariseDescription } from '../../scripts/scripts.js';
+import { formatDateUTCSeconds, isGatedResource, summariseDescription } from '../../scripts/scripts.js';
 import {
   a, div, h3, p, i, span,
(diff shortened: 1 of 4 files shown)


blocks/card/card.js
```diff
@@ -62,6 +62,7 @@ class Card {
     this.c2aLinkStyle = false;
     this.c2aLinkConfig = false;
     this.c2aLinkIconFull = false;
+    this.showDate = true;
```
andreituicu, line 65: I think this should be `false` by default, otherwise you display the date everywhere the card class is being used.
for example, if you look at the following two places, the date is now added there too and I don't think you want that:
* https://latest-resources--moleculardevices--hlxsites.hlx.page/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers#related-products-services
* https://latest-resources--moleculardevices--hlxsites.hlx.page/lab-notes/blog

then you will override the default `false` value, only in the latest-resources carousel code here:
https://github.com/hlxsites/moleculardevices/blob/main/blocks/latest-resources/latest-resources.js#L62-L65

andreituicu (commented):
The PR title doesn't match the code, making it quite confusing. It says `latest resources carousel sorting`, whereas the code doesn't do any changes for sorting, it is adds dates to cards/carousels.

dev-rajneeshkumar:
> The PR title doesn't match the code, making it quite confusing. It says `latest resources carousel sorting`, whereas the code doesn't do any changes for sorting, it is adds dates to cards/carousels.

Actually this was also part of requirement which has been fixed by sorting query index sheet only so changes removed from code

blocks/card/card.js
dev-rajneeshkumar: Thanks for pointing this out @andreituicu , this should be captured in unit testing but missed. We have fixed it can you please review again?

andreituicu (approved):

PR #1298, opened 2024-01-19
loading issue and speed optimization

dev-rajneeshkumar:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.live/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers
- After: https://moldev-opt--moleculardevices--hlxsites.hlx.live/products/microplate-readers/multi-mode-readers/spectramax-i3x-readers

### DIFF
--- a/blocks/header/helpers.js
+++ b/blocks/header/helpers.js
@@ -1,3 +1,4 @@
+/* eslint-disable linebreak-style */
 import { getMetadata } from '../../scripts/lib-franklin.js';
 import {
   a, div, li,
@@ -50,7 +51,7 @@ export function expandMenu(element) {
   element.setAttribute('aria-expanded', 'true');
 }
 
-export function buildBrandLogo(content) {
+export async function buildBrandLogo(content) {
   const logoImg = content.querySelector('.nav-brand');
 
   const logoLink = a(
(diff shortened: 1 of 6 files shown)


blocks/header/helpers.js
```diff
@@ -50,7 +51,7 @@ export function expandMenu(element) {
   element.setAttribute('aria-expanded', 'true');
 }
 
-export function buildBrandLogo(content) {
+export async function buildBrandLogo(content) {
```
mhaack, line 54: This is not doing much I wonder how making it async and `await buildBrandLogo(content)` should improve any performance?

blocks/latest-resources/latest-resources.js
```diff
@@ -29,7 +30,6 @@ async function getResourcesFromMetaTags() {
 
   return ffetch('/query-index.json')
     .sheet('resources')
-    .chunks(2000)
```
duynguyen, line 32: @dev-rajneeshkumar just to make sure you are aware of this change: with 2k chunk size if will only take 1 call to fetch all resources to run the filter (at this point of time with 1.9k resources). with default chunk size (255) it takes 8 calls for the same number of resources. And in the end, the filter will run through all 1.9k resources.

mhaack: Plus we habe HTTP/2 should not matter mucht.

mhaack (approved):

blocks/latest-resources/latest-resources.js
dev-rajneeshkumar: ok @duynguyen let see the performance and user experience after this PR merge and then will change if required

PR #1300, opened 2024-01-24
set events chronological order

dev-rajneeshkumar:
Set order of featured events as chronological order. Currently it's as per created date

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/newsroom
- After: https://featured-events--moleculardevices--hlxsites.hlx.page/newsroom


mhaack (approved):

blocks/latest-events/latest-events.js
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable linebreak-style */
```
mhaack, line 1: I'm not a fan of this as this can be easily avoided

PR #1301, opened 2024-01-24
have multiple faq blocks on a page with one schema

dev-rajneeshkumar:
Have multiple FAQ blocks on a page using single block and single schema to avoid multiple schemas issue in google.

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--moleculardevices--hlxsites.hlx.page/products/gxp-compliance-solutions/softmax-pro-gxp-software/gxp-assessment-questionnaire
- After: https://faq-block--moleculardevices--hlxsites.hlx.page/products/gxp-compliance-solutions/softmax-pro-gxp-software/gxp-assessment-questionnaire

### DIFF
--- a/blocks/accordion/accordion.js
+++ b/blocks/accordion/accordion.js
@@ -1,3 +1,4 @@
+/* eslint-disable linebreak-style */
 import { decorateIcons } from '../../scripts/lib-franklin.js';
 import {
   div, span,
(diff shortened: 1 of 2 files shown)


blocks/accordion/accordion.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable linebreak-style */
```
mhaack, line 1: Again, lets remove this. Either disable the rule globally or better do not breakt it. Does not make much sense to add this line everywhere.

mhaack (approved):

